Issue #3, opened 2023-11-21
Allow creating of a Windows window


BrentFarris:
This does not need to be a fully feature complete window, just one that opens, closes and has the events setup and ready to go.

BrentFarris:
Pull request video:  https://youtu.be/LGzxv_FOd58

PR #4, opened 2023-11-21
#3 Allow creating of a Windows window

BrentFarris:
Using the win32 API to create a window and read it's events. This uses a novel way of communicating between C and Go using a shared pointer.

https://youtu.be/LGzxv_FOd58

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,5 +1,9 @@
 package main
 
+import (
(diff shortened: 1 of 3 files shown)


BrentFarris:
Pull request video:  https://youtu.be/LGzxv_FOd58

Issue #17, opened 2023-11-22
Create code for showing an X11 window on Linux


BrentFarris:
This should follow the windows code. This should also refactor the windows code structure to separate C code and share go code.

BrentFarris:
Pull request video:  https://youtu.be/jd4StXSAEcM

PR #18, opened 2023-11-22
#17 Create code for showing an X11 window on Linux

BrentFarris:
Refactoring the MS Windows code to make it able to share code between platforms. Also this adds in the X11 code for presenting a window on Linux

https://youtu.be/jd4StXSAEcM


BrentFarris:
Pull request video:  https://youtu.be/jd4StXSAEcM

Issue #6, opened 2023-11-21
Create transform as part of matrix package


BrentFarris:
This should include basic transformations as well as support for parenting transforms to one another

BrentFarris:
Pull request video:  https://youtu.be/X8UP4nuxjmc

PR #19, opened 2023-11-23
#6 Create transform as part of matrix package

BrentFarris:
Transform created and allows for parenting of transforms. This does not include Quaternion for rotation yet.

https://youtu.be/X8UP4nuxjmc


BrentFarris:
Pull request video:  https://youtu.be/X8UP4nuxjmc

Issue #7, opened 2023-11-21
Create the host (mediator) and entity


BrentFarris:
Host and Entity are core concepts to the engine, so they will be created simultaneously here. The host is the mediator to all systems, and the entity is the identifier to an instance of a transform and other data. Entities are housed within the host mediator.

BrentFarris:
Pull request video:  https://youtu.be/Kc1pBJW1Pjk

PR #21, opened 2023-11-23
#7 Create the host (mediator) and entity

BrentFarris:
This is the start of both the Host and the Entity. Both are not fully complete yet, but are basically functional in their current state. I also moved the window code into the host so that it wasn't floating around in the main. I will need to move the window code into it's own container as we can have multiple windows.

https://youtu.be/Kc1pBJW1Pjk

### DIFF
--- a/engine/entity.go
+++ b/engine/entity.go
@@ -1,0 +1,295 @@
+package engine
+
+import (
+	"kaiju/matrix"
+)
+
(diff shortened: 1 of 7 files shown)


BrentFarris:
Pull request video:  https://youtu.be/Kc1pBJW1Pjk

Issue #9, opened 2023-11-21
Create the primary logic for mouse input


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/xIWre47U8nI

PR #23, opened 2023-11-23
#9 Create the primary logic for mouse input

BrentFarris:
This creates the complete logic for mouse input to be used. This also modifies the window to have a mouse contained within it. Lastly this hooks up the (windows only) mouse input events for the mouse

https://youtu.be/xIWre47U8nI

### DIFF
--- a/hid/mouse.go
+++ b/hid/mouse.go
@@ -1,0 +1,142 @@
+package hid
+
+import (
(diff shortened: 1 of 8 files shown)


BrentFarris:
Pull request video:  https://youtu.be/xIWre47U8nI

Issue #22, opened 2023-11-23
Hook up the mouse input events for X11


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/kfATaVIVU18

PR #24, opened 2023-11-23
#22 Hook up the mouse input events for X11

BrentFarris:
This is working X11 mouse input events. There is also a bit of refactor for the windows mouse events here to make more common code between the windowing systems. I refactored the windows code so that which button was being clicked is being sent rather than just X1 and X2


BrentFarris:
Pull request video:  https://youtu.be/kfATaVIVU18

Issue #26, opened 2023-11-25
Create project management scripts when creating a new project


BrentFarris:
- Create a go mod file for the project
- Create a build batch/shell script for the project
- Create a `.vscode` folder with launch Windows/X11 script

BrentFarris:
Pull request video:  https://youtu.be/RbTg29VvVCI

PR #28, opened 2023-11-25
#26 Create project management scripts

BrentFarris:
Create build scripts, project management scripts, and also bootstrap code. This moves the bootstrapping to it's own folder and properly sets up the source.go to use the engine.Host without circular references. This also creates the go.mod file and .vscode launch scripts.


BrentFarris:
Pull request video:  https://youtu.be/RbTg29VvVCI

Issue #15, opened 2023-11-21
Setup yaegi Go interpreter


BrentFarris:
[Yaegi](https://github.com/KaijuEngine/yaegi) is a Go interpreter. When it loads up it should `Eval` all go code within the `source` folder of the project. The `[project]/source.Main` function shall be called with the system host pointer.

BrentFarris:
Pull request video:  https://youtu.be/QOn3r2nD1N0

PR #29, opened 2023-11-26
#15 Setup yaegi Go interpreter

BrentFarris:
This is a functional implementation of the yaegi Go interpreter for interpreting Go code as a scripting language for the engine. This is built so that when the release build of the game/application is built, the go code is compiled directly into the executable rather than being interpreted. I've also added the code needed to execute the project source code at any time on a given host.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@
 
 # Go workspace file
 go.work
+
+project_template/
+bin/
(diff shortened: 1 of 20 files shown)


BrentFarris:
Pull request video:  https://youtu.be/QOn3r2nD1N0

Issue #30, opened 2023-11-26
Use project name for build scripts on new project


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/pWTvZyJuX4Q

PR #31, opened 2023-11-26
#30 Use project name for build scripts

BrentFarris:
This now uses [PROJECT_NAME] string replacement to replace with the project name. Also as a part of this commit I'm adding the build scripts for the editor itself by just copying the existing project build scripts and adding the editor tag and the kaiju name.


BrentFarris:
Pull request video:  https://youtu.be/pWTvZyJuX4Q

Issue #33, opened 2023-11-27
Cleanup initial Win32 window code


BrentFarris:
- Move message handling to `window_proc`
- Select shared memory by hwnd
  - Can use `SetWindowLongPtrA` and `SetWindowLongPtrA`
    - `SetWindowLongPtrA(hwnd, GWLP_USERDATA, (LONG_PTR)myPtr);`
    - `LONG_PTR usrData = GetWindowLongPtrA(hwnd, GWLP_USERDATA);`
    - Note that the first couple results from `GetWindowLongPtrA` will be `NULL`
- Create a struct to hold shared memory and size
- Write fatal error message string to shared memory
- Error on fatal error from window construction

BrentFarris:
Pull request video:  https://youtu.be/RORrQXRaEA0

PR #34, opened 2023-11-28
#33 Cleanup initial Win32 window code

BrentFarris:
Adding a fatal message to be read from the caller of the window creation. If a fatal is returned from window create, then it will panic with the message. I've also better wrapped the shared memory into the InputEvent structure so that it can be written to without the need of memcpy. I also updated X11 to use the helper methods created for the shared memory.


BrentFarris:
Pull request video:  https://youtu.be/RORrQXRaEA0

Issue #32, opened 2023-11-27
Get a very basic OpenGL implementation started


BrentFarris:
This should just initialize OpenGL and clear the screen on both the Win32 and X11 windows.

BrentFarris:
Pull request video:  https://youtu.be/p_BZSwumQAY

PR #35, opened 2023-11-28
#32 Get a very basic OpenGL implementation started

BrentFarris:
Adding the code for both Windows and Linux. I also updated the build scripts to include the libraries needed for compiling both Win32 and X11 for use of OpenGL. This required Glad and Glad_wgl on Windows so I copied that from one of my C game engines and also had to generate the wgl header from the glad generator. This is just doing a very basic OpenGL clear screen to cornflower blue


BrentFarris:
Pull request video:  https://youtu.be/p_BZSwumQAY

PR #39, opened 2023-11-30
Shader load & triangle draw in OpenGL on Win32/X11

BrentFarris:
This creates the base concept of a shader for OpenGL. It also creates a shader cache to hold shaders that have been built for reuse. This also creates the stub for all GL renderering through a C wrapper layer. Lastly a triangle is crudely drawn on the screen to verify the shader load/compile/use.

### DIFF
--- a/assets/database.go
+++ b/assets/database.go
@@ -1,0 +1,14 @@
+package assets
+
+import "kaiju/filesystem"
+
+type Database struct {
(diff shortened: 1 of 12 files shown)


BrentFarris:
Pull request video:  https://youtu.be/Q64SmEvy7xQ

Issue #40, opened 2023-11-30
Create the standard vertex layout for rendering


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/qK2WkGPTlLw

PR #41, opened 2023-11-30
#40 Create standard vertex layout for rendering

BrentFarris:
This creates a Vertex structure so and the associated code to load it up for rendering. The renderer is modified to have the create mesh function. Meshes are bare minimum right now as the focus was on the vertex layout. I also added Color, Color8, and Vec4i helper structs into the matrix library.

### DIFF
--- a/content/basic.frag
+++ b/content/basic.frag
@@ -1,6 +1,8 @@
 #version 330 core
 out vec4 FragColor;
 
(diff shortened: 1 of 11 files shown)


BrentFarris:
Pull request video:  https://youtu.be/qK2WkGPTlLw

Issue #42, opened 2023-11-30
Create perspective, orthographic, and turntable cameras


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/9GghKSeLHnM

PR #46, opened 2023-12-04
#42 Create perspective and orthographic cameras

BrentFarris:
Working perspective and orthographic camera code. This needed to add Ray and Frustum into a collision package, which also required helper functions and planes to be described. I also exposed Rad2Deg and Deg2Rad. Then I added the OPENGL tag to the build and debugging scripts. I then modified the shaders to have global data containing the camera matrices. Lastly I set the global data for the shader to the pre-calculated values from the camera code. There was a change to make the window shared memory into a pointer rather than embedded into the structure to fix a Go memory pinning issue that came up from the Renderer having pointer functions as well.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -10,7 +10,7 @@
 			"request": "launch",
 			"mode": "auto",
 			"program": "${workspaceFolder}",
-			"buildFlags": "-tags editor",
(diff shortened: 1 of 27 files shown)


BrentFarris:
Pull request video:  https://youtu.be/9GghKSeLHnM

Issue #47, opened 2023-12-07
Make window spin lock relinquish time back to OS


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/1WGhzGQ98pw

PR #48, opened 2023-12-08
#47 Make window spin relinquish time back to OS

BrentFarris:
On windows we need to set the priority of the currently running thread to idle for the window events while it waits on the main loop to continue. This lowers the check-in calls from the OS to this thread as it's setting idle and spinning. We then re-stack the thread on both X11 and Windows if it is not ready to read yet. If it is ready to read, we return the thread priority back to what it was (normal). I also fix the long-running GL crash as I need the program to run longer for testing idle CPU usage.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,8 +6,13 @@ import (
 	"kaiju/gl"
 	"kaiju/matrix"
 	"kaiju/rendering"
+	"runtime"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


BrentFarris:
Pull request video:  https://youtu.be/1WGhzGQ98pw

Issue #49, opened 2023-12-11
Cleanup rendering and make instanced rendering default


BrentFarris:
Currently the rendering code is set int the `main.go` file for testing. I want to move the rendering off to the correct location and call it as I would in production code. This will also require me to refactor how the rending is handled and change it to instanced rendering.

BrentFarris:
Pull request video:  https://youtu.be/GnddikJEBAs

PR #53, opened 2023-12-13
#49 Cleanup rendering, make instanced draw default

BrentFarris:
This changes the sample code in main to use instanced drawing and instanced data controlled through an instance group. This is not a complete finish of the implementation as we'll need to share shaders and only bind the shader once. This will likely be through a shader draw which then has all the instance groups beneath it. For now the instancing and writing of instance data into a texture for GL to use in the shader is functional.

### DIFF
--- a/content/basic.vert
+++ b/content/basic.vert
@@ -1,4 +1,5 @@
 #version 330 core
+
 layout (location = 0) in vec3 Position;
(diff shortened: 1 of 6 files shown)


BrentFarris:
Pull request video:  https://youtu.be/GnddikJEBAs

Issue #52, opened 2023-12-13
Create a shader render housing draw instance groups


BrentFarris:
We only want to bind the shader once and then draw all instance groups after it is bound, this way we're not re-binding an already bound pipeline or re-binding it later after another one is bound (out of order). This should basically just be a structure with a shader pointer and slice of draw instance groups. This also means that the shader can be removed from the draw instance groups.

BrentFarris:
Pull request video:  https://youtu.be/50W4TxY6lhI

Issue #51, opened 2023-12-13
Cache GL instance data texture, update pixels each frame


BrentFarris:
Currently we are generating and deleting a texture every frame for instance data. This change will generate/regenerate the texture as instances are added to the group and delete the texture when all instances are removed or the group is destroyed.

BrentFarris:
Pull request video:  https://youtu.be/50W4TxY6lhI

PR #54, opened 2023-12-13
#51 and #52 Improve GL instance drawing

BrentFarris:
This makes it so that shaders are only bound once and will draw all of the instances that are bound do them. This also fixes an issue of re-generating the texture every frame for draw instance groups instance data. The texture is now re-crated only when the group expands.


BrentFarris:
Pull request video:  https://youtu.be/50W4TxY6lhI

Issue #62, opened 2023-12-15
Setup textures and the texture cache


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/xbCL1xzdBsw

PR #63, opened 2023-12-15
#62 Setup textures and the texture cache

BrentFarris:
Textures are now working, though the full management of them isn't complete. Also how they are being drawn as part of the instance group may be refactored, but for now it is working. The part that isn't complete is getting the texture location by the string name in the shader for OpenGL. This prepares the engine to work with PNG, ASTC, and raw textures. This also sets up the texture cache for re-using textures


BrentFarris:
Pull request video:  https://youtu.be/xbCL1xzdBsw

PR #66, opened 2023-12-18
Adding some missing things I need for a port

BrentFarris:
Adding directory helper functions, file helper functions, and color helper functions. I'm also fixing the project template export for `.vscode` as it was missing linking and tags. Also added bitmap set of functions to make it easier to work with a mapping of bits.


BrentFarris:
Pull request video:  https://youtu.be/nyQVJM4P22k

Issue #65, opened 2023-12-18
Further push rendering code to be more streamline


BrentFarris:
Something needs to manage all of the drawings that need to happen. This thing should also make it a little easier for adding things to be rendered.

BrentFarris:
Pull request video:  https://youtu.be/n7N5RUPcMhQ

PR #67, opened 2023-12-18
#65 Further push rendering code to be simpler

BrentFarris:
Moving some of the rendering code around to prepare the rendering of meshes to be a little easier to do.


BrentFarris:
Pull request video:  https://youtu.be/n7N5RUPcMhQ

Issue #68, opened 2023-12-18
Prepare mesh code for loading mesh data


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/5x51frrMNXE

PR #69, opened 2023-12-18
#68 Prepare mesh code for loading mesh data

BrentFarris:
Adding in mesh cache for loading meshes by key. Updated main sample and tested code that was made in this and previous PRs. Added helper method for setting the model on the instance data. Added lots of mesh generation functions for primitive types of meshes. Added skipping of drawing instances if the target mesh is not ready to be drawn. Fixed bugs in caches where they weren't actually caching. Also removed old texture flipping as it isn't needed anymore.


BrentFarris:
Pull request video:  https://youtu.be/5x51frrMNXE

Issue #70, opened 2023-12-19
Further simplify drawing requests to a single structure


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/3j_2gydh8v8

Issue #50, opened 2023-12-13
Fix rotation issue found during issue #49


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/3j_2gydh8v8

PR #71, opened 2023-12-19
#50 & #70 Rot fix & drawing simplification

BrentFarris:
Fixing the issue with the rotation matrix causing issues, it was an issue with the generated matrix multiplication code. Also made it so that you can request something to be drawn using a single helper structure rather than going through multiple functions for grouping of drawings. This one structure is used to figure out what grouping the drawing belongs to and places it within it. If no group is found, then the group is created.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,30 +17,41 @@ type TriangleShaderData struct {
 	Color matrix.Color
 }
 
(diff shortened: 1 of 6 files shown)


BrentFarris:
Pull request video:  https://youtu.be/3j_2gydh8v8

Issue #58, opened 2023-12-13
Setup MSDF font rendering


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/vSk7HFXT9iY

PR #72, opened 2023-12-20
#58 Setup MSDF font rendering

BrentFarris:
This is porting the code I've previously written for fonts in other engines. This is also porting the shaders I have used for those. I also have done a few modifications to clean up the code and make it easier to use. I've hidden away the caches in the host as well and made interfaceable accessor functions for them. Also I've updated the global shader data to include a UI camera as it will be used quite a bit for fonts and UI things. For now I've copied over all my pre-generated msdf fonts as well.


BrentFarris:
Pull request video:  https://youtu.be/vSk7HFXT9iY

Issue #55, opened 2023-12-13
Setup order independent transparency (OIT) in GL


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/Nutkfa3izCY

PR #73, opened 2023-12-21
#55 Setup OIT in GL (working pass)

BrentFarris:
Adding the first working implementation of OIT (order independent transparency) for GL. This will look very different in general from the Vulkan implementation in the future. I also resovled a viewport bug with the windows as X11 had the width/height hard-coded and windows didn't pass a width and height at all. Now the size is passed into the window creation process. Sub shaders is basically the OIT shader for now, that will likely need to be renamed and how the OIT shader is setup is a prime place for refactor/cleanup.  I have setup the shaders needed for OIT as well, they came from a previous engine so their conventions may need to be updated later. I've also updated the demo in main to demo the OIT render. Lastly, now separation of solid and transparent drawings is controlled through the UseBlending boolean found on the Drawing structure.


BrentFarris:
Pull request video:  https://youtu.be/Nutkfa3izCY

Issue #13, opened 2023-11-21
Create the primary logic for cursor input container


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/QEoO3IFnZz8

Issue #12, opened 2023-11-21
Create the primary logic for stylus input


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/QEoO3IFnZz8

Issue #11, opened 2023-11-21
Create the primary logic for touch input


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/QEoO3IFnZz8

PR #74, opened 2023-12-24
#11 #12 #13 Stylus, touch, and cursor

BrentFarris:
In this one I've stubbed in the needed code to get touch and stylus input working abstractly. This code will need to be used on the various platforms to get the input events and push them through to the input system still though. I also created a "cursor" which is a wrapper around mouse/touch/stylus as a cursor (or pointer) input device. This will make it a little more interchangeable in the long run to not be checking specific cursor input devices.


BrentFarris:
Pull request video:  https://youtu.be/QEoO3IFnZz8

Issue #75, opened 2023-12-27
Create the base UI, layout, and grouping code


BrentFarris:

BrentFarris:
Pull request video: https://youtu.be/EeKtmhaxm-E

PR #77, opened 2023-12-28
#75 Create the base UI, layout, and grouping code

BrentFarris:
I have added all of the base ui code. This includes the grouping to help with ui elements that overlap that are related, it prevents things like click-through UI. I also have added the layout code which has all the math to do layouts for floating and stretched UI. There is a small addition to bitmap so that it has a new true option that will make a bitmap with all bits set to 1. The events for the UI have also been setup in here for requesting execution through the UI itself (if there isn't a group) or through the group. Some of the layout code may not be used, but it available for things like like CSS/HTML designed UI.

### DIFF
--- a/bitmap/bitmap.go
+++ b/bitmap/bitmap.go
@@ -8,6 +8,14 @@ func New(length int) Bitmap {
 	return make([]byte, LengthFor(length))
 }
 
+func NewTrue(length int) Bitmap {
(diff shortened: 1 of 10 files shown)


BrentFarris:
Pull request video: https://youtu.be/EeKtmhaxm-E

Issue #78, opened 2024-01-20
Create the panel ui


BrentFarris:

BrentFarris:
Pull request video:  https://youtu.be/ORM6xrGMMRw

PR #81, opened 2024-01-20
#78 Create the panel ui

BrentFarris:
The primary focus of this pull request was to create the "panel" ui. This is the primitive (other than label) that will be used to compose other more complex UI elements. This also gives the first test run of the layout system to make sure it is working properly. I did find a bug with the title bar being over the viewport on windows #79.

I also did some extra cleanup here to add in helper color functions, fix transformation multiplication on the shader data models, and also clean up the transforms after the render. I also fixed an issue where the texture width/height wasn't set until it was loaded on the GPU, so now that is set earlier.


BrentFarris:
Pull request video:  https://youtu.be/ORM6xrGMMRw

PR #510, opened 2025-12-10
docs: Specify windows-amd64 Go distribution requirement

Rudra-rps:
Fixes #48 - Updated documentation to explicitly mention installing the 64-bit (windows-amd64) version of Go instead of the 32-bit (windows-386) version, which causes Vulkan compilation errors.

Changes:
- README.md: Added prominent note about 64-bit requirement
- build_from_source.md: Enhanced Windows Requirements section with download link and verification steps
- build_from_source-custom.md: Already had correct documentation

Users can now verify their Go architecture with 'go env GOARCH' (should output 'amd64').


BrentFarris:
Thank you, more docs and clarity is always welcome 👍

PR #511, opened 2025-12-10
engine/ui: Fix potential nil pointer dereference

pierrec:
Stumbled on this while reading the ui implementation.


BrentFarris:
👍 Thanks!

PR #530, opened 2025-12-13
Add Alt+right-click smooth zoom for stage camera

### LINKED ISSUE #495: Add Alt+right-click to smooth zoom
In the stage view, you can use the mouse scroll to zoom, but I'd like to add a smooth zoom where you can hold Alt+right-click and then move the mouse horizontally/vertically to zoom based on the mouse's delta distance.

IrvingMg:
Adds Alt+right-click drag as an alternative zoom method for the stage view camera.

- Hold Alt + right-click and drag to zoom
- Drag up/right to zoom in, down/left to zoom out
- Works in both 3D and 2D modes

Closes #495

### DIFF
--- a/src/editor/editor_controls/editor_camera.go
+++ b/src/editor/editor_controls/editor_camera.go
@@ -214,7 +214,8 @@ func (e *EditorCamera) update3d(host *engine.Host, _ float64) (changed bool) {
 	if kb.HasAlt() || kb.KeyHeld(hid.KeyboardKeySpace) {
 		changed = true
 	}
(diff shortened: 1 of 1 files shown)


BrentFarris (approved):
👍 The zoom scale is a little bit aggressive, but the function looks to be working. I'll merge this and update the zoom scale for this case.

Thank you!

IrvingMg:
@BrentFarris BTW, I noticed there's no UI displaying the current zoom level. Would it be useful to add a small indicator showing the zoom in the stage view? 

Happy to open an issue and work on it if interested.

BrentFarris:
@IrvingMg, that's okay. I think we can leave this to just be what you see in the editor without adding other UI elements. Thank you for the suggestion though!

PR #540, opened 2025-12-14
editor(#528): add undoable remove action for entity data bindings

### LINKED ISSUE #528: Add "X" (delete) button for entity data binding
1. Open a project
2. Go to stage workspace
3. Create an entity
4. Attach an entity data to the entity

There should be a button in the UI (possibly just an "X" in the corner) to remove the data binding from the entity. This action should be added to the editor's history, so a developer should be able to undo/redo the action.

Take care to ensure all the fields in the entity data are properly restored when undoing. The easiest way to do this would likely be to just detach the data from the entity and store it's pointer in the history memento. Then when re-attaching, just put the pointer back on.

rajanarahul93:
## Summary
Added a delete ('X') button to the entity details panel allowing developers to remove attached entity data bindings. The action integrates with the editor's history system for full undo/redo support.

## Root cause
Entity data bindings could be attached but had no removal mechanism in the UI.

## Changes
- Added `DetachDataBinding` and `AttachDataBinding` methods to `StageEntity`.
- Created `EntityDataDetachHistory` struct to handle undo/redo by storing the data pointer.
- Added "X" button to the Entity Details UI that triggers the detach history action.

## Testing
- Validated that clicking "X" removes the binding.
- Validated that Undo (Ctrl+Z) restores the binding with all data fields intact.
- Validated that Redo (Ctrl+Y) removes it again.

## Related
Closes #528


src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
```diff
@@ -104,6 +104,7 @@ func (dui *WorkspaceDetailsUI) setupFuncs() map[string]func(*document.Element) {
 		"searchEntityData":  dui.searchEntityData,
 		"addEntityData":     dui.addEntityData,
 		"changeData":        dui.changeData,
+		"removeEntityData":  dui.removeEntityData,
```
BrentFarris: This `removeEntityData` should be defined in the HTML (see my other comment for details).

BrentFarris (dismissed):
This has the base of what is needed, I will need to make some edits in post though. Thanks!

Issue #416, opened 2025-11-16
Escape should close (cancel) overlays


BrentFarris:

Quantum-Fiend:
Hi, I’d like to work on issue #416 (“Escape should close overlays”)—could you please confirm if I can take it on?

BrentFarris:
Sure thing, thanks @Quantum-Fiend! Give it a shot and let me know your thoughts. It may be that we can consolidate some of the overlay startup code. Right now it's very linear and non-consolidated.

Quantum-Fiend:
@BrentFarris 

I’ve reviewed the new_project_overlay and noticed that it doesn’t have explicit handling for the Escape key to close the overlay. I understand that you want to prevent the Escape key from closing this specific overlay.

I plan to make the following changes:

- Modify the file_browser overlay to respect a new flag (EscapeCloses) that controls whether Escape can close the overlay.

- In new_project_overlay, pass the EscapeCloses: false flag when invoking the file browser, ensuring that it cannot be closed by the Escape key.

Before proceeding, I wanted to confirm if this approach aligns with your expectations for the behavior of the new_project_overlay. Please let me know if you’d prefer a different approach or if there’s anything I might have missed....

Thank you for your time and guidance!

BrentFarris:
So `file_browser` already has the escape key built in, it's one of the few that had it setup. However, I don't believe all the other overlays have the escape key support:
```
ai_prompt
confirm_prompt
context_menu
create_entity_data
input_prompt
reference_viewer
table_of_contents_overlay
```

PR #541, opened 2025-12-14
Prevent Escape from closing new_project_overlay

Quantum-Fiend:
This change prevents new_project_overlay from being closed via the Escape key.

Escape is intercepted locally at the overlay level and consumed, ensuring the
overlay remains open. The keyboard callback is properly removed on overlay
destruction.

This matches the exception described in issue #416.

Fixes : #416


BrentFarris:
The issue #416 is actually asking for escape to cancel/close transient overlays.

Issue #537, opened 2025-12-14
Persist Editor/Project Settings


bshore:
**Is your feature request related to a problem? Please describe.**
Editor should save some common high-level information.
- 2D / 3D Camera mode
- Most recent stage or default (`LaunchParams.StartStage`)
- Editor window size, monitor placement (~fancy)

**Describe the solution you'd like**
```js
// settings.json
{
  "RecentProjects": {
    "/foo/bar/baz": {
      "CameraMode":"2D/3D",
      "Stage":"{stage-identifier}"
    },
  },
  "EditorWindow": {
      "Width":944,
      "Height":500,
      "Monitor": "Default/Dell/ViewSonic/etc..??"
  }
}
```

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

bshore:
I'll look into this one next

bshore:
I think this is the action plan, probably in multiple phases:

## Phase 1

I've landed on storing `CameraMode` and `LatestOpenStage` within Project Settings. I think this feels right because if someone has multiple projects say a 2D and a 3D game, when that project is opened up in the Editor it should correctly default the stage camera mode. The same is true for `LatestOpenStage`, whatever was most recently opened per project will be saved to Project Settings.

In order to do this, Project Settings will need to be treated similarly to Editor Settings. Clicking 2D/3D or opening a stage needs to be able to update a Project's `settings.json` so when/if the Editor is closed, the next time it's opened it persists `CameraMode` and `LatestOpenStage`

## Phase 2

Expose or "bubble up" window resize events from Vulkan up through the Window and to the Editor so the Width/Height can be persisted. I'm not quite sure how to implement this just yet, but I've been digging through:
- `src/bootstrap/bootstrap_game.go` -- `bootstrapLoop(...)` -- particularly `container.Run(...)`:
  - `src/engine/host_container/host_container.go` -- `Run(width, height, x, y int, platformState any)`
    - `src/engine/host.go` -- `Initialize(width, height, x, y int, platformState any)`
- `src/platform/windowing/window.vk.go`
- `src/rendering/vk_instance.go`
- `src/rendering/vk_swap_chain.go`

This phase should enable persisting window size. Saving the config file should be limited to a "mouse release" event if possible.

## Phase 3

[...]

PR #552, opened 2025-12-18
#537 phase 1, persist camera mode and latest stage

bshore:


src/editor/editor_workspace/stage_workspace/stage_workspace_ui_data.go
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************/
-/* stage_workspace_editor_interface.go                                        */
```
bshore: GH diff being weird. This was a file delete & file add.

Interface was moved to `src/editor/editor_workspace/editor_workspace.go` - more able to be referenced without `import cycle not allowed` - I think this needs a second pass

bshore:
@BrentFarris I was able to get around the cyclical import by defining another interface. Review when you have time, happy to address any feedback

BrentFarris:
This PR helped me find a bug I needed to fix:
https://github.com/KaijuEngine/kaiju/commit/d87e250ab5d6dc448123d51c8640ae30d99555fe

Also, there was a small refactor to account for gizmo attachments:
https://github.com/KaijuEngine/kaiju/commit/7ef5f3c59c1403c118021a1f1a00c073ba2d6225

Issue #139, opened 2024-02-06
Add auto-triangulation to the obj loader


BrentFarris:
If you try to load with quads, it'll be all messed up. See the [code for the quads](https://github.com/KaijuEngine/kaiju/blob/master/src/rendering/loaders/obj_loader.go#L143-L153)

DAShoe1:
I'd like to tackle this issue. I have a solution ready but wanted to make sure there aren't any particular expectations for the implementation. I've added simple fan triangulation and added some basic error handling.

BrentFarris:
@DAShoe1 Thanks for the interest! That's all fine. I've not used the OBJ loader in a while, so hopefully it is all working otherwise.

PR #556, opened 2025-12-19
Improved support for importing OBJ files

DAShoe1:
This PR resolves issue #139.

This PR also fixes support for multiple objects and splits objects that have multiple materials into separate meshes.

I also added the ability for the editor to import OBJ files for testing.


BrentFarris:
Awesome, thank you! I'm working on something around content and meshes, so I'm going to merge this one in before I break it with some changes. Thank you for the contribution!

PR #581, opened 2025-12-29
GLTF import all mesh primitives

DAShoe1:
This is a straight forward change that just takes the existing import logic and has it iterate over each primitive instead of just the first primitive of each mesh.


BrentFarris:
👍 Thanks, I'll try and check this out tomorrow. I spent all night tonight fighting with transformation issues (which I fixed), so I haven't had much time for anything else.

BrentFarris (approved):
Sweet, thanks! 👍

PR #597, opened 2026-01-09
events: add Manager event system for inter-entity signaling

Akif-jpg:
- Add new package engine/systems/events with:
  - Manager, Channel, Subscription, EventData types
  - Async (Emit) and sync (EmitSync) delivery methods
  - Thread-safe subscription management using RWMutex
  - Utilities: CreateChannel, GetChannel, RemoveChannel, Clear
- Add unit tests covering basic behavior, concurrency, types and benchmarks (src/engine/systems/events/event_manager_test.go)
- Add examples demonstrating entity communication, host integration and channel management (src/engine/systems/events/event_manager_example_test.go). Examples use EmitSync where deterministic output is required.
- Integrate event manager into Host:
  - add eventManager *events.Manager field
  - initialize in NewHost and expose via EventManager() getter
  - clear manager during Teardown
- All new comments are in English and follow project style.

This commit implements a publish-subscribe event system intended for lightweight, in-process signaling between entities and systems.

### DIFF
--- a/src/engine/host.go
+++ b/src/engine/host.go
@@ -131,6 +131,7 @@ type Host struct {
 	LateUpdater      Updater
 	assetDatabase    assets.Database
 	physics          StagePhysics
+	eventManager     *events.Manager
 	OnClose          events.Event
 	CloseSignal      chan struct{}
 	frameRateLimit   *time.Ticker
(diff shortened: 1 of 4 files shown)


BrentFarris (commented):
Sorry about the delayed response here, I've not ignored or forgotten about this pull request. I've just been a bit tied up in some other features I've been working on. Thank you for the PR!

Adding in a pub/sub is a bit of an architecture change for things like entity destruction, but it does simplify a lot of other types of events. I can see it being useful as a `framework` tool for game developers. I'll try and make some time to think more about something like this being in architecture or as a helper for developers.

Issue #492, opened 2025-12-09
Add press-and-hold down on delete key in input box to repeat delete characters


BrentFarris:
1. Navigate to input box
2. Type some text
3. Press and hold delete key

Currently this deletes 1 character and nothing more. The goal would be that as the key is held down, after a short delay, it begins continuously deleting characters

muratmirgun:
Hello, I can take this issue/

BrentFarris:
@muratmirgun Excellent! Thank you for the interest, assigned

PR #609, opened 2026-01-28
feat: press and hold backspace to delete continously

relaxgameing:
With reference to #492

Changes Made:
1. Changed keyboard struct to store time of last button press for each key (when the state changes from idle -> down)
>  The reason for adding this is because there was no way of knowing how long key was held for so having this gives us access to that.
2. Updated input keyPressed function for press and hold for continuous delete

Here is a demo video
https://github.com/user-attachments/assets/734701fc-6ada-4f55-8331-5469443c16bc

**Note: only tested for mac**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,6 @@ kaiju_android.so
 
 # GoLand
 .idea/
+
+#Mac
+.DS_Store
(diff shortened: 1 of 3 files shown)


BrentFarris:
👍 Awesome, thanks! I'll take a look at this tomorrow

relaxgameing:
> 👍 Looks good, thank you!

Still Learning 😅

PR #611, opened 2026-01-30
Add right-click copy for log entries (#455)

### LINKED ISSUE #455: Add right-click copy of an entry in the log

HoraDomu:
Fixes #455 — added the ability to right click a log entry in the status bar
and copy its full text to the clipboard.


ghost:
Mistake pull

PR #612, opened 2026-01-30
Fix: remove example addLogEntry calls, keep right-click copy function…

HoraDomu:
Removed the hardcoded example log entries from the status bar HTML. The addLogEntry function remains functional, but the log popup no longer auto populates with test messages. This fixes the issue of unnecessary logs appearing in the editor UI. Fixing my own mistake.


ghost:
I had made other pull request on accident with a draft code this is the correct fix for issue #455

BrentFarris:
The engine doesn't support JavaScript

Issue #569, opened 2025-12-23
Add copy/paste of transform, shader data, and entity data


BrentFarris:
As a developer I should be able to right-click "Copy" on an entity transform, shader data, or entity data. I then should be able to right-click "Paste Values" or "Paste as New". As an extended goal, it would be nice to paste specific fields (like "Scale").

- [x] Setup framework by getting it working with transform (position, rotation, scale)
- [ ] Use framework on shader and entity data

relaxgameing:
Hi @BrentFarris , can I give this issue a shot?

BrentFarris:
> Hi [@BrentFarris](https://github.com/BrentFarris) , can I give this issue a shot?

Sure thing! Thanks, @relaxgameing

Note that I already setup the base copy/paste for entity data. It only creates new copies of the entity data though, so it doesn't do the value pasting option.

relaxgameing:
Hi @BrentFarris , I have made a pull request for the issue, it's not complete (shader data is remaining) but the main idea is implemented can you review it once and see if it fits your convention 😅

BrentFarris:
Thanks for the ping @relaxgameing, I'll check it out :)

BrentFarris:
The framework for this has been setup (merged) and is functional on the transform position/rotation/scale. Next steps would be to use the framework for the generated shader data fields and the generated entity data fields.

PR #613, opened 2026-02-01
feat: Copy and Paste entity attributes inside details panel

relaxgameing:
**With reference to :  #569** 

Now we can copy and paste selected entity transform data 

Here is a Demo Video

demo
**the video quality is low because of 10MB size limit**

**Note: I have not implemented all the copy paste feature only the transform attributes** 
> Reason is to have the code checked out  for logic and implementation detail before implementing further as i have very minimal understanding of the kaiju architecture workflow and structure

### DIFF
--- a/src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
+++ b/src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
@@ -44,6 +44,7 @@ import (
 	"kaiju/editor/editor_overlay/context_menu"
 	"kaiju/editor/editor_stage_manager"
 	"kaiju/editor/editor_stage_manager/data_binding_renderer"
+	"kaiju/editor/memento"
 	"kaiju/engine"
 	"kaiju/engine/ui"
 	"kaiju/engine/ui/markup/document"
@@ -639,11 +640,18 @@ func (dui *WorkspaceDetailsUI) commonChangeData(e *document.Element, isShaderDat
 		return false
 	}
 	entity := sel[0]
-	h := &detailsDataChangeHistory{DetailsWorkspace: dui}
+	h := &detailsDataChangeHistory{}
 	success := false
 	if isShaderData {
 		v := reflect.ValueOf(entity.StageData.ShaderData).Elem().Field(idx)
-		h.Value = v
+		h.ValueChangeProcedure = func(newVal reflect.Value) {
+			v.Set(newVal)
+			if reload, ok := dui.TargetedElementValueReload[v]; ok {
+				reload()
+			} else {
+				dui.reload()
+			}
+		}
 		h.From = v.Interface()
 		if success = reflectAssignChanges(e, v); success {
 			h.To = v.Interface()
@@ -653,7 +661,14 @@ func (dui *WorkspaceDetailsUI) commonChangeData(e *document.Element, isShaderDat
 		if err == nil {
 			target := entity.DataBindings()[pIdx]
 			v := reflect.ValueOf(target.BoundData).Elem().Field(idx)
-			h.Value = v
+			h.ValueChangeProcedure = func(newVal reflect.Value) {
+				v.Set(newVal)
+				if reload, ok := dui.TargetedElementValueReload[v]; ok {
+					reload()
+				} else {
+					dui.reload()
+				}
+			}
 			h.From = v.Interface()
 			if ok := reflectAssignChanges(e, v); ok {
 				data_binding_renderer.Updated(target, weak.Make(w.Host), entity)
@@ -1015,10 +1030,10 @@ func toFloat(str string) float64 {
 // document.Element must have a `data-copypaste` attribute which would represent The
 // target attribute which need's to be copied
 func (dui *WorkspaceDetailsUI) copyAttribute(e *document.Element) {
-	defer tracing.NewRegion("WorkspaceDetailsUI.copyEntityPosition").End()
+	defer tracing.NewRegion("WorkspaceDetailsUI.copyAttribute").End()
 	w := dui.workspace.Value()
 	if w == nil {
-		slog.Error("WorkspaceDetailsUI:copyEntityPosition workspace ptr is nil")
+		slog.Error("WorkspaceDetailsUI.copyAttribute workspace ptr is nil")
 		return
 	}
 	sel := w.stageView.Manager().Selection()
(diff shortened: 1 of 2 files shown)


relaxgameing:
Now i have also noticed that the html file is changed fully because of my vs code formatter , i would change it in future update to this branch

relaxgameing:
I have changed the logic of the copy and pasting and tried to make it more generic where we are using the `data-copy` and 
`data-paste`, data fields in the html to identify which field we are copying into the `copiedAttribute` field and where we need to paste the value stored in the `copiedAttribute` according to `data-copy` and `data-paste` fields 

the main idea behind this was to make a generic way where we copy any value and paste it into any field where the type is same

BrentFarris:
👍 Thanks! Making those more generic is great, it should be a guide in copying other fields in the future.
Were you wanting to resolve your HTML formatting stuff as part of this PR?

BrentFarris (approved):

relaxgameing:
> Were you wanting to resolve your HTML formatting stuff as part of this PR?

No actually, i changed the html to add buttons for copy and paste  and I use a vscode extension  named 'Prettier'
So that automatically formatted the html ( this was not intended if you want i could try reverting it back , I have not changed much just added few btns and little bit of css). 

And actually I made this pull request so that i could get your review about the methodology before moving further, cause I have implemented it with copy and paste btn ( before implementing the right-click drop-down way of doing it as described in the issue) , so i will add new commits in this branch after adding the drop down way of pasting then for shader data

BrentFarris:
> No actually, i changed the html to add buttons for copy and paste and I use a vscode extension named 'Prettier' So that automatically formatted the html ( this was not intended if you want i could try reverting it back , I have not changed much just added few btns and little bit of css).

Okay, cool. Yes, if it's just a couple of changes (adding the attribute, and `onrightclick`), then having just those changes would be preferred so that the HTML is consistent with the other files.

> And actually I made this pull request so that i could get your review about the methodology before moving further, cause I have implemented it with copy and paste btn ( before implementing the right-click drop-down way of doing it as described in the issue) , so i will add new commits in this branch after adding the drop down way of pasting then for shader data

You're right! I'm very sorry, I was actually playing a game with my uncle earlier when I checked the review and forgot about the right-click part and focused mostly on the reflection part of the code you had.

Over all, I think doing the reflection is the best way for it to scale later.

That being said (though I don't want to add too much onto this issue), I was thinking about it a bit further; I wonder if there is some clean up that needs to be done so that we could identify the target value from it's parent container. Say, something like `id="position"` on the parent `div`, then in code we can map the string `"position"` to a reflected value. In this way, if we were to right-click on the parent and do "copy", it could just reference the `reflect.Value` associated to that id. Then if we do a right-click "paste", it could set the `reflect.Value` for that target to the referenced one.

This would also open up the possibility to gray out the "paste" on fields that don't match in kind/type.

I don't want to go out of scope for what you're doing here on that, but if it's something you'd be interested in trying out in this PR, feel free. Of course, all this PR would need to focus on is the hook-up for position/rotation/scale. I can then expand that entity data and shader data fields following that framework.

So, in the HTML, we would do something like `id="transformPosition"` on the parent div (that holds the X/Y/Z input boxes). Then in the Go code we could make a function, or mapping to associate that string (`transformPosition` in this case) to a `reflect.Value` that holds the copied value. This would be re-used for the paste to set the value through reflection to the previously copied value.

What do you think?

relaxgameing:
> So, in the HTML, we would do something like `id="transformPosition"` on the parent div (that holds the X/Y/Z input boxes). Then in the Go code we could make a function, or mapping to associate that string (`transformPosition` in this case) to a `reflect.Value` that holds the copied value. This would be re-used for the paste to set the value through reflection to the previously copied value.
> 
> What do you think?

I think it should be achievable right now. to do so just add the `data-copy="<typeof copyPasteKind>"` on the parent div then when we right click and select copy it would then run the `copyAttribute` function there we could get the string value of the `data-copy` and then update the `copiedAttribute` in dui and do the same with the paste 

so it would require minimal changes of just how you get the `data-copy/paste` from the dropdown overlay or so ( i have to check that)

relaxgameing:
@BrentFarris , I have made some changes and now it is working with the dropdown here is a demo video:
i have not changed the concept much just little bit of changes and i have also corrected the html file formatting thing  so now it is showing only the changed part


https://github.com/user-attachments/assets/55aa7b6c-eefe-477d-b953-62a9a6b5dc6e

src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
```diff
@@ -980,3 +1010,122 @@ func toFloat(str string) float64 {
...
+	}
+
+	slog.Info("entity", "field to copy", fieldToCopy, "copied", dui.copiedAttribute)
+}
+
+// Prerequisite:
+// document.Element must have a `data-copypaste` attribute which would represent The
+// target attribute where it needs to be pasted
+func (dui *WorkspaceDetailsUI) pasteAttribute(e *document.Element) {
+	defer tracing.NewRegion("WorkspaceDetailsUI.pasteEntityData").End()
```
BrentFarris: `WorkspaceDetailsUI.pasteEntityData` -> `WorkspaceDetailsUI.pasteAttribute`

src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
```diff
@@ -980,3 +1010,122 @@ func toFloat(str string) float64 {
...
+			sel[0].Transform.SetPosition(dui.copiedAttribute.(matrix.Vec3))
+		}
+	case entityRotation:
+		if fieldMap[pasteToField] == reflect.TypeOf(sel[0].Transform.Rotation()).Kind() {
+			sel[0].Transform.SetRotation(dui.copiedAttribute.(matrix.Vec3))
+		}
+	case entityScale:
+		if fieldMap[pasteToField] == reflect.TypeOf(sel[0].Transform.Scale()).Kind() {
+			sel[0].Transform.SetScale(dui.copiedAttribute.(matrix.Vec3))
+		}
```
BrentFarris: We should also have a history entry for transform changes, I believe there should be code for this when you change the value of the input box that you can pull from.

src/editor/editor_workspace/stage_workspace/stage_workspace_details_ui.go
```diff
@@ -980,3 +1010,122 @@ func toFloat(str string) float64 {
...
+		options = append(options, context_menu.ContextMenuOption{
+			Label: fmt.Sprintf("Copy %s Attribute", valToCopy),
+			Call:  func() { dui.copyAttribute(e) },
+		})
+	}
+
+	if dui.copiedAttribute != nil {
+		kind, ok := copyPasteStrToKindMap[valToCopy]
+		if !ok {
+			slog.Info(fmt.Sprintf("%s is not a registered type", valToCopy))
```
BrentFarris: Could be turned into a `Warning` rather than info. Also might need some details of where this log is coming from.

relaxgameing:
The framework is ready to be merged but implementation for shader and entity data is still missing so it's your call now 😅

BrentFarris:
> The framework is ready to be merged but implementation for shader and entity data is still missing so it's your call now 😅

I'm cool with that, I like smaller incremental changes. Splitting a single task up into two or more PRs makes it easier to read and work with in my opinion (even if later changes refactor the implementation). It's also good to have this split between the two because transform and shader/entity data are presented differently (due to shader/entity data fields being generated), this may cause some other refactors that will be easier to review in isolation.

Issue #619, opened 2026-02-07
When a panel/list is updated, the scroll will keep it's offset


BrentFarris:
This is a harmless bug overall, but it does cause some pretty confusing UI. Notice in this video that when I click on a filter, and the list is updated to only have 2 items, it looks blank. Once I scroll the mouse wheel, the panel then gets correctly updated to scroll to the correct position.

https://github.com/user-attachments/assets/dbfed6cb-70cf-4b6f-8af8-5f7d1cc587f3

This is a systemic problem, basically the scroll needs to update when children are added/removed from a panel.

relaxgameing:
@BrentFarris , Can i Have it ?

BrentFarris:
> @BrentFarris , Can i Have it ?
> 

Sure thing, thanks! I've assigned it.

I don't know why I assigned this to me rather than unassigned, probably out of habit.

Issue #351, opened 2025-10-28
After final tag usage is removed (no more usages), it should be removed from stage and content workspace


BrentFarris:

BrentFarris:
Reposting ([this message](https://github.com/KaijuEngine/kaiju/pull/620#issuecomment-3872310328)) here as well

---

Currently, the way that this is done is through [editor events](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_events/editor_events.go#L41). This is the preferred method at the moment due to them being easily traceable and debugable using the symbol lookup of the code editor. This event is already exposed in both [stage](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/stage_workspace/stage_workspace_editor_interface.go) and [content](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/content_workspace/content_workspace_editor_interface.go#L51) workspaces. Creating an event that the content workspace can execute should be enough for those ones. The content workspace can execute the event when a new tag is generated or the last usage of the tag is removed.

Let me know if you thinks this works, or if you have a better idea for it. I've not used Redux for React, so I'm not sure what all that covers. I'm assuming it's something like a pub/sub?

PR #620, opened 2026-02-08
Fix: #619 content preview scroll reset after filter change

relaxgameing:
Reseting the content preview ui panel scroll when ever there is a change in the filter
> not running this in mainThread because the scroll is not resetting instantly but waiting for ui update caused by any ui event 

Demo:

https://github.com/user-attachments/assets/089af349-a3dd-46f0-9d27-5eeed6a65799

### DIFF
--- a/src/editor/editor_workspace/content_workspace/content_workspace.go
+++ b/src/editor/editor_workspace/content_workspace/content_workspace.go
@@ -806,6 +806,7 @@ func (w *ContentWorkspace) runFilter() {
 			e.UI.Hide()
 		}
 	}
+	w.contentList.UIPanel.ResetScroll()
 	w.Host.RunOnMainThread(w.Doc.Clean)
 }
 


BrentFarris:
👍 Thanks, there is a similar issue with the content workspace if you don't mind having a fix for that, as part of this PR.

I, unfortunately, have duplicated code in stage and content workspaces for the content views. This is because they are somewhat the same, but not exactly the same.

relaxgameing:
Sure 😁, should I also include issues: #350 and #351 ( tags related issue)

BrentFarris:
> Sure 😁, should I also include issues: #350 and #351 ( tags related issue)

Sure thing! That would be awesome, thanks!

relaxgameing:
i have pushed only the scroll fix for both the stage and content workspace 

> Sure 😁, should I also include issues: #350 and #351 ( tags related issue)

for these two issue i have looked into the `editor.go` and how much i understood is that there is no inter-workspace communication which is causing states to become stale. so to solve this i might have to find an elegant way for inter-workspace communication ( i am thinking about redux , which is a react state management lib , we could get that kind of pattern or something simpler) and i would create a different PR for that

( if there is a way already present for the communication then please inform 😅 and it means i have not read the code properly🥲)

Sorry for the inconvenience

BrentFarris:
> i have pushed only the scroll fix for both the stage and content workspace

Awesome, thank you!

> ( if there is a way already present for the communication then please inform 😅 and it means i have not read the code properly🥲)

Currently, the way that this is done is through [editor events](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_events/editor_events.go#L41). This is the preferred method at the moment due to them being easily traceable and debugable using the symbol lookup of the code editor. This event is already exposed in both [stage](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/stage_workspace/stage_workspace_editor_interface.go) and [content](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/content_workspace/content_workspace_editor_interface.go#L51) workspaces. Creating an event that the content workspace can execute should be enough for those ones. The content workspace can execute the event when a new tag is generated or the last usage of the tag is removed.

Let me know if you thinks this works, or if you have a better idea for it. I've not used Redux for React, so I'm not sure what all that covers. I'm assuming it's something like a pub/sub?

relaxgameing:
> Currently, the way that this is done is through [editor events](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_events/editor_events.go#L41). This is the preferred method at the moment due to them being easily traceable and debugable using the symbol lookup of the code editor. This event is already exposed in both [stage](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/stage_workspace/stage_workspace_editor_interface.go) and [content](https://github.com/KaijuEngine/kaiju/blob/master/src/editor/editor_workspace/content_workspace/content_workspace_editor_interface.go#L51) workspaces. Creating an event that the content workspace can execute should be enough for those ones. The content workspace can execute the event when a new tag is generated or the last usage of the tag is removed.

I saw the editor events feature  but the thought didnt came into my mind 😂 , might be because i was rushing. but now that i think about this i think this is a nice way of passing events and can have all the events in one place and just register for events which we require 

> Let me know if you thinks this works, or if you have a better idea for it. 
 
Yes, i think this might be the best way to keep things simple .

>I've not used Redux for React, so I'm not sure what all that covers. I'm assuming it's something like a pub/sub?

Yes, generally these state management ways are pub/sub , in redux what we do is create a store which has all different reducers (different state which deals with different thing) and each reducers has some events (we define those events) which you can dispatch with the new intended state from any part of the component tree and after each dispatch every component which are using the specific value that has been updated will re-render and all the components children's also re-render.

here in the editor we have to manually update the html ui every time we are changing the states , but i react what we do is we just define how should a particular component looks like and we could just change the parameter which that component uses can it automatically changes its ui , same way here we are dependent on the cache contents , so when ever cache content changes the ui should reflect that automatically. Only thing that i think is missing is ui re-rendering when ever pageData changes then we could by the help of the editor event make ui changes more event driven rather than manually making changes on the dom ourself

PR #621, opened 2026-02-10
fix: Tag state is now stable btw content and stage workspace

relaxgameing:
With Reference to: #350 & #351 

Demo

https://github.com/user-attachments/assets/635dd7b0-caca-4012-a798-6acee369a48d

This is the most intuitive and direct implementation which uses the existing structure of the editor, but it feels kinda a hacky to me because we need to change so many things each time and have to maintain the state manually which could cause  us to make mistakes. ( need you opinion on this)

### Improvements
1. i am thinking of making some changes in workspace structure such as switching the pageData.Tags from `[]string` -> `map[string]int` as it would be more faster and better approach  for lookup and for getting the  count of  number of entities using a particular tag previously we are looping 
2. there should be some ways to re-render ui's like this :
```go
func (cui *WorkspaceContentUI) handleNewFilterTag(newTag string) {
    slog.Info("New Tag recieved")
    w := cui.workspace.Value()
    w.pageData.Tags = append(w.pageData.Tags, newTag)
    cui.filterArea.UI.SetDirty(ui.DirtyTypeGenerated)
    cui.filterArea.UI.Clean()
}
```
as it is more declarative , cleaner and better in my opinion.
3. i am thinking of adding a `div` at the very last of the html file to include all the templates which might be required in that html the div will be hidden so it doesn't affect the ui but is accessible to us for spawning new entities of that template ( as i have used a very unefficient way of spawning a new tag filter btn in stage workspace

### DIFF
--- a/src/editor/editor_workspace/content_workspace/content_workspace.go
+++ b/src/editor/editor_workspace/content_workspace/content_workspace.go
@@ -43,6 +43,7 @@ import (
 	"kaiju/editor/editor_overlay/confirm_prompt"
 	"kaiju/editor/editor_overlay/context_menu"
 	"kaiju/editor/editor_overlay/file_browser"
+	"kaiju/editor/editor_overlay/input_prompt"
 	"kaiju/editor/editor_workspace/common_workspace"
 	"kaiju/editor/project/project_database/content_database"
 	"kaiju/editor/project/project_file_system"
@@ -631,7 +632,18 @@ func (w *ContentWorkspace) clickDelete(*document.Element) {
 	})
 }
 
+func (w *ContentWorkspace) removeContentView(id []string) {
+	defer tracing.NewRegion("ContentWorkspace.removeContentView").End()
+	for i := range id {
+		if elm, ok := w.Doc.GetElementById(id[i]); ok {
+			w.Doc.RemoveElementWithoutApplyStyles(elm)
+		}
+	}
+	w.Doc.ApplyStyles()
+}
+
 func (w *ContentWorkspace) completeDeleteOfSelectedContent() {
+	defer tracing.NewRegion("ContentWorkspace.completeDeleteOfSelectedContent").End()
 	ids := w.selectedIds()
 	for _, id := range ids {
 		err := content_database.Delete(id, w.pfs, w.cache)
@@ -707,6 +719,10 @@ func (w *ContentWorkspace) rightClickContent(e *document.Element) {
 		{
 			Label: "Create table of contents",
 			Call:  w.requestCreateTableOfContents,
+		},
+		{
+			Label: "Change GUID",
+			Call:  func() { w.requestChangeGuid(id) },
 		},
 	}
 	if cc, err := w.cache.Read(id); err == nil {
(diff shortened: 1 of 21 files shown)


BrentFarris:
Sweet thanks! Ill check this out tomorrow morning.

relaxgameing:
In this new commit i have changed the implementation of the tags/type filters from slices to map's which allows us to remove a good amount of iterations 

during this process i have also found a bug . when we add same tag string multiple times in the input box or by clicking on the suggestions then it makes the state invalid, so in next commit the objective is to fix it

src/editor/editor_workspace/content_workspace/content_workspace.go
```diff
@@ -505,6 +515,9 @@ func (w *ContentWorkspace) clickDeleteTag(e *document.Element) {
 	if !found {
 		slog.Error("failed to locate the tag that was expected to exist", "tag", tag)
 	}
+
+	w.editor.Events().OnTagRemoved.Execute(tag)
```
BrentFarris: The `clickDeleteTag` function is when we're deleting a single tag from the selected content. The tag may still be on content that we don't have selected (other content). So, would this call to the event cause the tag to be removed from the UI prematurely in that case?

Example:
We have 3 content, A, B, and C. All 3 have the tag "testing". If I select A and B, then click delete on the tag, C still has the "testing" tag, so it shouldn't yet be removed globally.

relaxgameing:
> I know you've still got plans to update this PR, but I wanted to review before going to bed.
> 
> First of all, thank you for being proactive about cleaning things up! I have thrown a lot of UIs together to get broader features tested/working and the content workspace (and stage workspace duplication) has been a bit of a sore point. So, any additional cleanup is much appreciated.
> 
> Second, I have a couple comments in this review, please forgive me if I missunderstood the `OnNewTagAdded` and `OnTagRemoved`. It is past midnight right now and my tired eyes may be deceiving me.

@BrentFarris , I have resolved all you concerns from the previous review and now this should be working fine as it was intented.
i would still suggest that before merging you should run it and test it out once because there was so many things that changed so there is chance of edge cases where ui stales or so.

one more thing i have notices is that in the cache content for the tags we can use the Set type which i have introduced in this PR so that we could remove a good amount of iterations which are there in the code base for checking if the content has a tag or not, and we could also change the Cache struct by merging the lookup and cache fields into a single map of `map[string]CacheContent` ( if there is any reason for having them as separate slices (cache localization optimization) then we could leave them as it is)

BrentFarris (approved):
Seems to be working for me, thanks! I'll keep an eye out for any issues related to this change.

Also to answer the question on slices, they were something that was never refactored, so I'm happy with the `set` change.

relaxgameing:
> Seems to be working for me, thanks! I'll keep an eye out for any issues related to this change.
> 
> Also to answer the question on slices, they were something that was never refactored, so I'm happy with the `set` change.

In this PR i didn't touched the cache or cache content struct but there we could actually do some refactor to reduce the access and loopup complexity, should I do it ?

BrentFarris:
> In this PR i didn't touched the cache or cache content struct but there we could actually do some refactor to reduce the access and loopup complexity, should I do it ?

Sure, I would be interested to see what you have in mind.

relaxgameing:
> > In this PR i didn't touched the cache or cache content struct but there we could actually do some refactor to reduce the access and loopup complexity, should I do it ?
> 
> Sure, I would be interested to see what you have in mind.

Nothing crazy 😅, it would be a simple change.

PR #637, opened 2026-02-23
Add the full range of material icons as MSDF

### LINKED ISSUE #632: Generate the full range of material icons as MSFD
This is just updating the charset to do the full range of characters in the material icon font set. This _should_ be as simple as adding the range of characters into the charset.txt for the generator. Something like 0x0000-0x00FF would generate icons that have character code 0 through 255.

Do not just put an arbitrary large range like 0x00000000-0x1EFFFFFF because this will make the MSDF build take a very long time and most of the codes aren't even used. Try to narrow down to just the needed range to capture all of the icons.

mikenye:
Resolves #632


mikenye:
No problems! The texture was generated by the script in https://github.com/KaijuEngine/kaiju_binary_tools/pull/2.

BrentFarris:
Sorry for the delay, been busy with some vector stuff

PR #643, opened 2026-03-01
Mouse stutter fix and removal of flyCamFlickerFix

NilssonCreative:
Different solution to flyCamFlickerFix in src/editor/editor_controls/editor_camera.go. Cursor still not hidden (not sure if it's supposed to be) on right click.
Tested on Windows 11 with go build -tags="debug,editor" -o ../ ./


BrentFarris (dismissed):
👍 Sweet, thanks!

PR #645, opened 2026-03-02
Fixes editor crash on windows resize

NilssonCreative:
Fix to handle Suboptimal result from vk.AcquireNextImage as success.
Prevents the editor from crash when resizing, minimizing or maximizing the window.


BrentFarris:
👍 Thank you!

PR #653, opened 2026-03-22
fix: update overlay background color for better contrast

### LINKED ISSUE #652: The background color for the window is the same as the input box, making it look odd
1. Click on "Create" in the menu bar
2. Click on "Entity Data"

Image

The background color for the overlay should match other backgrounds with input boxes so that the box bounds is evident.

Deepak8858:
Update the background color of various overlays (Create Entity Data, Input Prompt, AI Prompt) from #121212 to #282828 to match other UI panels and provide better contrast for input boxes.

Fixes #652


BrentFarris:
Sorry, going to close this one as I've made an alternative fix using `edPanelBg`

PR #655, opened 2026-03-29
Feat/add texture cache to improve render performance

JrSchmidtt:
This PR introduces a **texture caching mechanism** into the rendering pipeline to reduce redundant resource creation and improve overall rendering performance.

The implementation aligns with the engine’s cache-oriented architecture, ensuring textures are reused instead of being recreated during runtime.

## Motivation

Previously, textures could be **loaded or instantiated multiple times** during rendering, especially in scenarios where the same asset is reused across entities or draw calls. This resulted in:

- Unnecessary CPU/GPU overhead  
- Increased memory usage  
- Performance degradation in texture-heavy scenes  

## Changes

- Added a **texture cache layer** to store and reuse previously loaded textures  
- Implemented cache lookup to avoid redundant texture creation  
- Integrated the cache transparently into the rendering flow  
- Reduced unnecessary texture allocation and upload operations  

## Impact

- Scope: Rendering system / resource management  
- Risk: Low  
- Compatibility: Non-breaking  

## How to test

1. Run a scene with multiple entities sharing the same texture  
2. Verify that textures are not recreated multiple times  
3. Monitor memory usage and rendering performance  
4. Compare behavior before and after the change  

## Notes

- Follows the engine’s existing caching patterns  
- Establishes a foundation for future improvements (e.g., cache eviction, reference counting)  
- No changes to the public API  

### Before (no cache)
https://www.youtube.com/watch?v=gqVPTZV1l44
image

cpu-no-cache

### After
https://www.youtube.com/watch?v=QgjHpqZcsuY
image

cpu

#

### DIFF
--- a/src/matrix/transform.go
+++ b/src/matrix/transform.go
@@ -95,7 +95,12 @@ func (t *Transform) Parent() *Transform  { return t.parent }
 func (t *Transform) Position() Vec3      { return t.relativePosition }
 
 func (t *Transform) IsDirty() bool {
-	return t.frameDirty
+	if !t.frameDirty {
+		return false
+	}
+	return !t.framePosition.Equals(t.position) ||
+		!t.frameRotation.Equals(t.rotation) ||
+		!t.frameScale.Equals(t.scale)
 }
 
 func (t *Transform) Right() Vec3 {


src/matrix/transform.go
```diff
@@ -95,12 +95,7 @@ func (t *Transform) Parent() *Transform  { return t.parent }
 func (t *Transform) Position() Vec3      { return t.relativePosition }
 
 func (t *Transform) IsDirty() bool {
-	if !t.frameDirty {
-		return false
-	}
-	return !t.framePosition.Equals(t.position) ||
-		!t.frameRotation.Equals(t.rotation) ||
-		!t.frameScale.Equals(t.scale)
```
BrentFarris: Perhaps an erroneous change?

This optimization is for transforms that change or possibly vibrate during the frame. In the case that a transform axis goes from 1 -> 0 then back from 0 -> 1 it will return that the transform is not dirty. This can happen quite often in the UI code during the tree stabilization and if not checked it could cause the stabilization to take longer or perhaps not even complete in cases of random vibration from style stabilization.

BrentFarris:
It seems that there may be an issue that I'm investigating from this merge that has texture width/height set to 0 (causing a Vulkan validation error). I should have a fix in tonight

BrentFarris:
This is the commit that reduces the complexity of the change, while still keeping the perf improvement. Please give this commit a try and see if your performance is still improved.

https://github.com/KaijuEngine/kaiju/commit/d40a900d795de2fc7e2e094ea0063cad88aa445d

PR #656, opened 2026-03-31
Detect X11 numpad key input + numlock check

Jsewill:
This might not be the best way to do this, but I wanted numpad keys to work as numbers. I gave claude a shot at fixing this, but I would have done it the same way it did.


BrentFarris (approved):
👍 Looks pretty straight forward, easy enough enhancement, thanks!

PR #657, opened 2026-03-31
Fixed missing SelectRoot() call in font rendering.

Jsewill:
I ran into an issue with font rendering. So, I figured I should make a pull request.


BrentFarris (approved):
👍 Great, thanks!

PR #689, opened 2026-04-12
feature/dynamic mesh and skip ui clean for inactive items.

Jsewill:
Some potential improvements, courtesy of my project and claude's insistence that there wasn't another, better way. My understanding of graphics libraries consists of older OpenGL standards, so please let me know if there is already a better way to do this in kaiju.

I ran into issues with frequent mesh updating and the GC, and needed something different than a DEVICE_LOCAL mesh. CreateDynamicMesh provides a HOST_VISIBLE type mesh, which allows for frequent updates.

Also, skipped cleaning the inactive UI tree items. I hope I didn't overlook something here, and that it isn't a problem that I also included this in the same PR.

### DIFF
--- a/src/engine/ui/ui.go
+++ b/src/engine/ui/ui.go
@@ -299,6 +299,9 @@ func (ui *UI) Clean() {
 	for !stabilized && maxIterations > 0 {
 		stabilized = true
 		for i := range tree {
+			if !tree[i].IsActive() {
+				continue
+			}
 			tree[i].cleanDirty()
 			tree[i].Layout().update()
 			tree[i].postLayoutUpdate()
@@ -307,6 +310,9 @@ func (ui *UI) Clean() {
 		maxIterations--
 	}
 	for i := range tree {
+		if !tree[i].IsActive() {
+			continue
+		}
 		tree[i].GenerateScissor()
 		tree[i].render()
 	}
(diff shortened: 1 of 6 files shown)


BrentFarris:
> Some potential improvements, courtesy of my project and claude's insistence that there wasn't another, better way. My understanding of graphics libraries consists of older OpenGL standards, so please let me know if there is already a better way to do this in kaiju.
> 
> I ran into issues with frequent mesh updating and the GC, and needed something different than a DEVICE_LOCAL mesh. CreateDynamicMesh provides a HOST_VISIBLE type mesh, which allows for frequent updates.

This is awesome, thank you! I've been hoping to get more into DMA stuff on the GPU, but I often wait until I have a solid need for features before adding them. I don't believe anything good comes out of speculative feature programming, someone needing the change typically has a good idea of what's required and what they need solved.

> Also, skipped cleaning the inactive UI tree items. I hope I didn't overlook something here,

No, this is fine, any branches of the tree items shouldn't need an update if they are not active themselves.

> and that it isn't a problem that I also included this in the same PR.

Nope! I'm not about process and bureaucracy. The only time I'd really like a separate PR is if there is a lot of work on 2 completely different ideas. These small QoL fixes in any PR are completely welcome!

BrentFarris (approved):
👍 Clean and straight forward implementation, thank you!

PR #690, opened 2026-04-12
Use meta key on macOS and enhance input navigation

Cronos87:
This PR is a proposition. It enables using the `meta` key on macOS and keep `ctrl` as it should be.

CMD+left on a input will move the cursor at the beginning, same behavior for the right.
ALT+left on a input will move the cursor to the previous break work, for the right, to the next break word.

Any comments or suggestions are very welcomed 🙏🏼

By the way, my VSCode reorder the import, I don't know if it is good or not.

### DIFF
--- a/src/editor/editor.go
+++ b/src/editor/editor.go
@@ -37,6 +37,9 @@
 package editor
 
 import (
+	"log/slog"
+	"time"
+
 	"kaijuengine.com/build"
 	"kaijuengine.com/editor/editor_embedded_content"
 	"kaijuengine.com/editor/editor_events"
@@ -64,8 +67,6 @@ import (
 	"kaijuengine.com/matrix"
 	"kaijuengine.com/platform/hid"
 	"kaijuengine.com/platform/profiler/tracing"
-	"log/slog"
-	"time"
 )
 
 // Editor is the entry point structure for the entire editor. It acts as the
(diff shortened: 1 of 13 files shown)


BrentFarris:
> This PR is a proposition. It enables using the `meta` key on macOS and keep `ctrl` as it should be.
> 
> CMD+left on a input will move the cursor at the beginning, same behavior for the right. ALT+left on a input will move the cursor to the previous break work, for the right, to the next break word.
> 
> Any comments or suggestions are very welcomed 🙏🏼

Great! This PR is mainly to bring Mac more in-line with the rest of Mac keyboard input ecosystem right? I don't use Mac a whole lot, but I know that Ctrl and Meta are often flipped from what Linux/Windows uses.

> By the way, my VSCode reorder the import, I don't know if it is good or not.

That's totally fine, a lot of files probably will wind up doing this in PRs simply because I renamed the repository package from "kaiju" to "kaijuengine.com" so that the `go doc` generator works correctly (it needs a "." in the package name to work).

Cronos87:
> Great! This PR is mainly to bring Mac more in-line with the rest of Mac keyboard input ecosystem right?

Precisely! At the moment, both `ctrl+c` and `cmd+c` will work on macOS, which is inconsistante. That's why I added the method `HasCtrlOrMeta()`, it will check the meta key on macOS only, not the ctrl key, and Windows/Linux will check ctrl key only.

For the methods `HasCtrl()` and `HasMeta()`, they do what they say independently of the platform.

> simply because I renamed the repository package from "kaiju" to "kaijuengine.com" so that the go doc generator works correctly

Understood! I won't hesitate to clean up the imports when I come across them :)

Thank you for the explanation and the fast answer!

BrentFarris (approved):
👍 Looks good, thanks for making Mac inputs more like Mac inputs. Also I like the clarity of `HasCtrlOrMeta`, it should make it clear that there is a platform split there without being overly verbose about it.

Issue #692, opened 2026-04-13
Texture Drag-and-Drop Applies Incorrect Texture Until Reload


Ahsan-Sarbaz:
**Bug Description**
Dragging a texture from the content browser to the stage works correctly for the first texture. However, when dragging a different texture afterward, the stage incorrectly applies the first texture again instead of the newly selected one. Interestingly, after saving and reloading the stage, the correct textures are restored.

**Steps to Reproduce**

1. Import multiple textures into the content browser
2. Drag a texture onto the stage — it is applied correctly
3. Drag a different texture onto the stage — the first texture is applied again instead of the new one

BrentFarris:
Thanks for the report, by the description it doesn't seem like it'd be too hard to work out.

Ahsan-Sarbaz:
I did send in a PR. https://github.com/KaijuEngine/kaiju/pull/693

BrentFarris:
Thanks!

PR #693, opened 2026-04-13
Create new key for textures on drag and drop if needed

Ahsan-Sarbaz:
Fix texture cache unique-key handling for inserted runtime textures

`InsertRawTexture` and `InsertImageTexture` were checking the cache before resolving `GenerateUniqueTextureKey` (`""`) into an actual UUID. That meant repeated calls intended to create fresh textures could reuse the same cache entry under the empty-string key, causing newly spawned images to display as copies of the first inserted texture.

This change normalizes the key with `selectKey(key)` before the cache lookup, so runtime-inserted textures get a real unique key prior to deduplication. That preserves normal cache reuse for explicit keys while making `GenerateUniqueTextureKey` behave as intended.


Ahsan-Sarbaz:
Fixes https://github.com/KaijuEngine/kaiju/issues/692

BrentFarris (approved):
Great, thank you!

PR #696, opened 2026-04-14
Center the window at launch on macOS

Cronos87:
At the moment, the window is placed at the bottom left of the screen when the engine is launched.

The project initialise the window at the coordinate -1, -1, which is valide on macOS.

This PR centers the window, following Apple style, which places the window centered horizontally, but not really vertically.


BrentFarris (approved):
Sweet, thank you! 👍

Cronos87:
Yeah, thank you 🎉

Issue #426, opened 2025-11-18
Strip all unreferenced content from the content archive


BrentFarris:
This should use the function created by #425 to go through each content and determine if it has any references.

xetra11:
@BrentFarris 
I can work on this. Assign please.

xetra11:
@BrentFarris 

Questions for clarification.

1. When saying that this task should strip all unreferenced content. Should it be...
  a. A button in the content view like "Purge" or "Tidy Up" that runs that functionality
  b. Only a function in code that will be bound to UI elements at a later stage
  c. \<Your Vision\>

BrentFarris:
> [@BrentFarris](https://github.com/BrentFarris)
> 
> Questions for clarification.
> 
> 1. When saying that this task should strip all unreferenced content. Should it be...
>    a. A button in the content view like "Purge" or "Tidy Up" that runs that functionality
>    b. Only a function in code that will be bound to UI elements at a later stage
>    c. <Your Vision>

Sorry for the lacking description on this issue. So my original intent for this issue was actually (C) - strip the content at built time. So, the asset database that is constructed for the built game does not have any unused assets.

I think it would be nice (later) to add a helper for purging unused assets from the project, or at least listing them for the developer to do cleanup. We can create a separate issue for that task though.

xetra11:
@BrentFarris 
Ah ok so you want a "tree-shaking" feature for the build. 
I will then try to hook it up to that.

PR #699, opened 2026-04-15
#426 add simple tree shake for content when building/packaging

xetra11:
WIP

_PS: I also added my GoLand Run Configuration for Debug mode in case anyone is working from the same IDE and wants a quick Editor Debug spinup_


xetra11:
resolves #426 

@BrentFarris 
Please tell me if the location of that logic is ok.
I was thinking about

PR #700, opened 2026-04-15
#426 Add logic to filter only referenced content from the content cache in the package step of the build pipeline

### LINKED ISSUE #426: Strip all unreferenced content from the content archive

xetra11:
resolves #426

### DIFF
--- a/src/editor/editor_common_workspace_interface.go
+++ b/src/editor/editor_common_workspace_interface.go
@@ -90,10 +90,6 @@ func (ed *Editor) ShadingWorkspace() *shading_workspace.ShadingWorkspace {
 	return &ed.workspaces.shading
 }
 
-func (ed *Editor) HasReferences() {
-
-}
-
 func (ed *Editor) ShowReferences(id string) {
 	ed.BlurInterface()
 	o, _ := reference_viewer.Show(ed.host, &ed.project, id)
--- a/src/editor/editor_common_workspace_interface.go
+++ b/src/editor/editor_common_workspace_interface.go
@@ -90,6 +90,10 @@ func (ed *Editor) ShadingWorkspace() *shading_workspace.ShadingWorkspace {
 	return &ed.workspaces.shading
 }
 
+func (ed *Editor) HasReferences() {
+
+}
+
 func (ed *Editor) ShowReferences(id string) {
 	ed.BlurInterface()
 	o, _ := reference_viewer.Show(ed.host, &ed.project, id)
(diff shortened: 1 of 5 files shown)


xetra11:
PS: I added a Run Config for Jetbrains GoLand that includes the config to run a Debug Editor. Just in case someone has a need for it.

src/editor/editor_common_workspace_interface.go
```diff
@@ -90,6 +90,10 @@ func (ed *Editor) ShadingWorkspace() *shading_workspace.ShadingWorkspace {
 	return &ed.workspaces.shading
 }
 
+func (ed *Editor) HasReferences() {
+
+}
+
```
BrentFarris: This empty function can be removed, or is there an intent to use this for something?

src/editor/project/project_references.go
```diff
@@ -67,6 +68,25 @@ func (p *Project) FindReferences(id string) ([]ContentReference, error) {
...
 
+// FindAllReferencedContentFromCache Identifies and returns a list of content entries from the cache that are
+// referenced in the project and therefore filtering out unreferenced content.
+func (p *Project) FindAllReferencedContentFromCache(cachedContent []content_database.CachedContent) []content_database.CachedContent {
+	var referencedContent []content_database.CachedContent
+
+	slog.Info("shake off all unreferenced content")
+	for _, content := range cachedContent {
+		_ = p.FindReferencesWithCallback(content.Id(), func(reference ContentReference) {
+			referencedContent = append(referencedContent, content)
```
BrentFarris, line 80: I believe that this will do an append each time a reference is found. So either it needs to be a unique append. Or you can use the `FindReferences` function and only add the content to the list if it's result has a `len` > 0. Though, this will go through every reference when it could just break out early (and add the content) after the first found reference.

BrentFarris (changes requested):

src/editor/editor_common_workspace_interface.go
xetra11: Woopsie - that's what reviews are for. That one slipped

src/editor/project/project_references.go
xetra11: Yeah I had that `FindReferences` with the len > 0 check before the `FindReference` + Callback flirted with me.
Since I only tested with a Stage reference I never realized it's duplicating entries.

Will fix asap

BrentFarris (approved):
👍 Looks good, thanks! I pulled and did some checking, but I'll merge this in and do a bit more testing.

PR #701, opened 2026-04-15
feat(editor): improve ReferenceViewer feedback for empty search results

### LINKED ISSUE #698: Add user facing info if no content references have been found
### Is your feature request related to a problem? Please describe.
When checking for references in the Content Browser, if an asset has no references, a blank popup is displayed. This creates confusion, as there is no indication of what the empty result means.

### Describe the solution you'd like
Instead of showing a blank popup, display a clear message (e.g., “No references found”) to inform the user that the asset does not have any references.

### Describe alternatives you've considered
None

### Additional context
This would improve usability and reduce confusion when working with assets that have no references.

Image

xetra11:
image

resolves #698


src/editor/editor_overlay/reference_viewer/reference_viewer_overlay.go
```diff
@@ -63,28 +64,29 @@ func Show(host *engine.Host, project *project.Project, id string) (*ReferenceVie
...
 	go func() {
-		if err := project.FindReferencesWithCallback(id, o.onFound); err != nil {
+		if err := project.FindReferencesWithCallback(id, referenceViewer.onFound); err != nil {
 			slog.Error("failed to find all references for content", "error", err)
 		}
-		if o == existing {
-			o.doc.GetElementsByClass("note")[0].UI.Hide()
+		if referenceViewer == existing {
+			referenceViewer.doc.GetElementsByClass("note")[0].UI.Hide()
+			referenceViewer.doc.GetElementsByClass("note-not-found")[0].UI.Show()
```
BrentFarris: Sweet, I think all we need is for this UI element to hide on startup, probably same time the template is hidden, otherwise it shows up when there are references.

Though, you may need to keep a count of found references (or a `boolean` flag) when a reference is found. That way you can call `Show()` only when the `count > 0` or that flag is set.

Image

BrentFarris (changes requested):

src/editor/editor_overlay/reference_viewer/reference_viewer_overlay.go
xetra11: Ugh sorry,
I was too much in a rush here. Will do better QA next time

xetra11:
Branch is too dirty

PR #706, opened 2026-04-19
fix(css): use compound selector for primary button styles

### LINKED ISSUE #703: Compound CSS hover event is overriding each of the compounds
This code should be making it so that only the `primary` button is turning red when hovered over. However, it is causing all buttons to be red when hovered. This is likely because the compound CSS statement `button` and `.primary` not being parsed correctly for events like `:hover`.

```css
button:hover, .button:hover {
	background-color: #575757;
}
button:active, .button:active {
	padding: 4px 5px 0px 5px;
	background-color: rgb(59, 59, 59);
}
button .primary, .button primary {
	background-color: #682A2D;
}
button .primary:hover, .button primary:hover {
	background-color: #881E1E;
}
```

https://github.com/user-attachments/assets/58c28458-f108-4152-a1c7-1c077ecf61f1

tmchow:
## Summary

Closes #703. The rule at `src/engine/ui/markup/css/default.css:426` is intended to style "a button that also has the `.primary` class", but it's written with a descendant combinator (space between `button` and `.primary`). In CSS, that instead matches `.primary` elements that are *inside* a button, which is not the intent.

As reported in the issue, the result is that the `:hover` rule on the descendant bleeds through to every button hover, so every button ends up red on hover rather than only the primary one.

## Fix

Use the compound-class form (no space between `button` and `.primary`):

```css
-button .primary, .button primary {
+button.primary, .button.primary {
    background-color: #682A2D;
}
-button .primary:hover, .button primary:hover {
+button.primary:hover, .button.primary:hover {
    background-color: #881E1E;
}
```

Also fixes the `.button primary` half of each pair, which was selecting an element with tag name `primary` inside `.button` (not what anyone wanted).

No other changes.

Closes #703

### DIFF
--- a/src/engine/ui/markup/css/default.css
+++ b/src/engine/ui/markup/css/default.css
@@ -423,10 +423,10 @@ button:active, .button:active {
 	/* background-image: url("input_atlas.png");
 	background-position: 64px 0; */
 }
-button .primary, .button primary {
+button.primary, .button.primary {
 	background-color: #682A2D;
 }
-button .primary:hover, .button primary:hover {
+button.primary:hover, .button.primary:hover {
 	background-color: #881E1E;
 }
 /** Unconventional helpers **/


BrentFarris:
There is more to this problem then moving the CSS around. If you run the debugger with your changes, you'll notice that it doesn't resolve the issue. Kaiju doesn't use a web view, there needs to be an update to the reader Go code to resolve this issue.

Please actually run and test your changes before you submit PRs. It wastes contributor time when you blindly submit PRs

PR #707, opened 2026-04-19
Fix: reset search filter input after adding selected entity data

xetra11:
In order to get access to the search field element I needed to invoke a weak pointer to `workspace`. However I don't feel that's the intended way to have access to another element inside an element action callback.
@BrentFarris 
Maybe you can direct me to a kaiju conventional way?

```go
entitySearchFilter, ok := dui.workspace.Value().Doc.GetElementById("entitySearchFilter")
if ok {
entitySearchFilter.UI.ToInput().SetText("")
}
```

### DIFF
--- a/src/editor/editor_embedded_content/editor_content/editor/ui/workspace/stage_workspace.go.html
+++ b/src/editor/editor_embedded_content/editor_content/editor/ui/workspace/stage_workspace.go.html
@@ -377,7 +377,7 @@
 						</div>
 					</div>
 					<div><span>Add data, or </span><button class="pasteIcon" onclick="pasteEntityDataAsNew">&#xE14F;</button></div>
-					<input type="text" placeholder="Find entity data..." onchange="searchEntityData" />
+					<input id="entitySearchFilter" type="text" placeholder="Find entity data..." onchange="searchEntityData" />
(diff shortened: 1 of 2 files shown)


BrentFarris:
> In order to get access to the search field element I needed to invoke a weak pointer to `workspace`. However I don't feel that's the intended way to have access to another element inside an element action callback.

The current flow, at the moment, for these UI elements is to store a reference to that UI element in the [setup function](https://github.com/KaijuEngine/kaiju/pull/707/changes#diff-0db81ec2ee6cf2e359a9198a4960f36e104b7ca44108fa8856ef91271637f519L161). Then you'll just have it as a member of the object and won't need to pull it from the DOM.

This adds another member, but that's fine since there is only ever 1 of this particular struct created.

xetra11:
@BrentFarris 
Sorry had a rough couple of weeks.
Can you tell me if this is still relevant without me checking through the codebase?

BrentFarris:
> @BrentFarris Sorry had a rough couple of weeks. Can you tell me if this is still relevant without me checking through the codebase?

This is no longer relevant as the way to attach entity data has changed. Thank you for your original contribution though!

PR #712, opened 2026-04-24
Windows dark titlebar

gitsomebit:
pr summary:
- adds dark titlebar support on windows
- changes window background color to match titlebar

### DIFF
--- a/src/platform/windowing/shared_mem.h
+++ b/src/platform/windowing/shared_mem.h
@@ -60,6 +60,7 @@ typedef struct {
 	RECT clientRect;
 	int mouseX;
 	int mouseY;
+	int titleBarMode;
 	bool rawInputFailed;
 	bool rawInputRequested;
 	struct {
--- a/src/platform/windowing/win32.c
+++ b/src/platform/windowing/win32.c
@@ -73,6 +73,8 @@
 #define WINDOW_TITLE_BAR_MODE_LIGHT 1
 #define WINDOW_TITLE_BAR_MODE_DARK 2
 
+static void apply_title_bar_mode(HWND hwnd, int mode);
+
 /*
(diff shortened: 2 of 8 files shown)


BrentFarris (approved):
Very nice! I just tested it out and it swaps with light/dark mode perfectly. It also gives us a starting point to hook into the color scheme switch for the UI as well.

Thank you!

gitsomebit:
just some follow up comments:
- developer can also enforce a color preference by setting titlebar mode during runtime
  ```go
  host.RunAfterTime(3*time.Second, func() {
    host.Window.SetTitleBarMode(windowing.TitleBarModeLight)
  })
  ```

code:
- I am not that familiar with `#cgo noescape` usage. Would be great if you look at that part just in case, but it should be fine
- was not sure where to store the titlebar mode for later retrieval. I settled on SharedMemory and am assuming you are fine with that. Will therfore add a getter for the developer to retrieve the mode value ~~(totally not forgot)~~

BrentFarris:
I saw those, being able to manually set is great.

The `#cgo noescape` tells the compiler that any pointers passed from Go to C will not escape the call (not stored in C)

The `#cgo nocallback` tells the compiler that the C function being called will not call back into Go at any point.

These two flags allow for the compiler to strip out some checks that it does for the C call and improves its performance.

Storing the current value inside of the shared memory is fine for now. Perhaps we can store the value and Go since you're not gonna get there any other way. Storing it in Go, will require us to call C at startup to get the current value, though.

gitsomebit:
added getter in #713

> Storing the current value inside of the shared memory is fine for now

I left the value in SharedMemory because its needed for OS theme changes to take effect. Added a comment to document this

PR #713, opened 2026-04-25
add titlebar mode getter

gitsomebit:
pr summary:
- add getter for getting titlebar mode
- add some valuable comments with previous implementation details


BrentFarris (approved):
👍 Nice clean-up, thanks!

PR #718, opened 2026-04-26
feat(window): add asserts

gitsomebit:
pr summary:
- add `Asserts` for windowing

NOTE: I mainly wanted to add an assert for validating `SetTitleBarMode` mode, but then I added some for the other functions as well


BrentFarris (approved):
👍 Thanks!

PR #719, opened 2026-04-26
feat(screen count): implement screen count on windows

gitsomebit:
pr summary:
- exposes screen count on windows
- other platforms have placeholders and always return `1`
- limitation is documented in comment of `MonitorCount()` comment

NOTE: this is intended as a pr before introducing multi-monitor positioning


BrentFarris (approved):
👍 Thanks!

I've added issue #720 for the TODOs in this PR.

PR #725, opened 2026-04-28
fix(titlebar): fix threading deadlock

gitsomebit:
pr summary:
- fix threading deadlock when using SetTitleBarMode() on worker thread

details:
using `SetTitleBarMode()` inside `func (EditorGame) Launch(host *engine.Host)` would deadlock because Launch is reached through `Host.Updater` (concurrent), so `host.Window.SetTitleBarMode(...)` can run on a worker thread. The Win32 implementation then called `SendMessage(WM_NCACTIVATE)` synchronouslya and the window thread is then waiting for the worker update to finish

### DIFF
--- a/src/platform/windowing/win32.c
+++ b/src/platform/windowing/win32.c
@@ -80,23 +80,24 @@ static bool user_prefers_dark_mode(void);
 * Messages defined here are NOT to be sent to other windows
 * https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerwindowmessagea#remarks
 */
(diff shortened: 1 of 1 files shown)


BrentFarris (approved):
👍 Excellent catch! Yes, all updates are run on worker threads. This is a design shift away from traditional game engines, but necessary for modern computers to take full advantage of all the logical threads.

There may be a couple more windows calls that would need to publish a message I'll need to take a look at. Setting the title bar text/icon are ones I can think of right now, so there are likely others.

Issue #722, opened 2026-04-28
Window updating is halted during window drag on Windows


BrentFarris:
https://github.com/user-attachments/assets/d07805c4-a385-4275-ab92-3a9b892c8ef5

BrentFarris:
Unfortunately, this is typical Winodows. The only way to avoid this, as far as I understand, is to create a borderless window with a custom title bar. That's a bit too much work at the moment, but may be something I can work out in the future

PR #729, opened 2026-04-29
File drop (windows)

gitsomebit:
pr summary:
- allow file-drops into editor for import
- supports dropping multiple files at once
- does `NOT` support folder drops yet

details:
- since file-drop is not something every application needs it is enabled with a build tag `filedrop`
- set up a router with handles that route processing
- in the editor drops are not allowed in project manager workspace and in all other workspaces we import all droped files int the content workspace (see video)
- drops are queued and processed during `Poll()` when file-drops are enabled (to ensure threading safety)

TODO: (future pr)
- allow dropping entire folders -> see #738

demo:

[demo](https://github.com/user-attachments/assets/f25ebb2e-bb62-4d43-8a80-6ea160f2d994)


src/engine/host.go
```diff
@@ -238,6 +239,16 @@ func (host *Host) InitializeRenderer() error {
...
+	if runtime.GOOS == "windows" {
+		wHost := weak.Make(host)
+		host.Window.OnMove.Add(func() {
+			host := wHost.Value()
+			if host == nil || host.Closing || host.Window == nil {
+				return
+			}
+			host.Render()
+		})
+	}
```
BrentFarris, line 251: Looks like this is an addition for issue #722 ?

src/editor/editor_file_drop_router.go
```diff
@@ -0,0 +1,162 @@
...
+	r.host = host
+	r.handleUnclaimedDrop = handleUnclaimedDrop
+	r.windowDropEventId = host.Window.OnFileDrop().Add(r.handleWindowFileDrop)
+}
+
+func (r *FileDropRouter) AddDropHandler(handler FileDropHandler) FileDropHandlerId {
+	defer tracing.NewRegion("FileDropRouter.AddDropHandler").End()
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	r.nextHandlerId++
```
BrentFarris, line 107: 👍 Thank you for incrementing before assignment so that `0` is an invalid id

BrentFarris (approved):
👍 Great stuff. I'm very excited about this one.

src/engine/host.go
BrentFarris: _Edit: I pulled down the PR to  test and it does at least fix the oddities with moving the window off screen (issue seen in this screenshot), thanks!_

image

BrentFarris:
I've also built the raw game to ensure everything is working correctly outside of the editor in a build 👍

src/engine/host.go
gitsomebit: > Looks like this is an addition for issue #722 ?

ups. looks like this one sliped through. I was experimenting with fixing #722, but only found this wanky workaround. It also does not fix window resize or updating during drag :(

I can remove it again, because I think the workaround is not good enough

PR #731, opened 2026-04-29
feat(content-workspace): add right-click option `open file in explorer`

gitsomebit:
pr summary:
- add right-click option when clicking on file in content workspace to open in explorer


BrentFarris (approved):
👍 Thanks!

PR #732, opened 2026-04-29
feat(css-units): add css units

gitsomebit:
pr summary:
- add missing css units
- use approximations based on current engine implementaion for `ch` and `rem`
- add tests to confirm correct units


BrentFarris (approved):
👍 Great, thank you for the tests as well

PR #736, opened 2026-04-30
Add View > Show/Hide Grid Toggle to Editor Menu

Colorfingers:
A simple change to the menu that adds a View tab with an option to Hide or Show the editor grid.

### DIFF
--- a/src/.claude/settings.local.json
+++ b/src/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(go build *)",
+      "Bash(echo \"build_exit=$?\")"
+    ]
+  }
+}
--- a/src/.claude/settings.local.json
+++ b/src/.claude/settings.local.json
@@ -1,8 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(go build *)",
-      "Bash(echo \"build_exit=$?\")"
-    ]
-  }
-}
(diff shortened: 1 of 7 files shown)


src/.claude/settings.local.json
BrentFarris, line 1: This file can be removed from this PR

Colorfingers:
I recommend you disapprove this pr.  The navigation does not work when the grid is hidden.  That's an unexpected outcome... I'll resubmit once I clean these things up.

BrentFarris:
> I recommend you disapprove this pr. The navigation does not work when the grid is hidden. That's an unexpected outcome... I'll resubmit once I clean these things up.

Oh interesting! I haven't tested this PR yet so I haven't seen that yet. Very strange that navigation is messed up from hiding the grid. I can look more into this later if you don't find anything obvious.

Colorfingers:
Actually, it does work... I was not trying it properly. Try it and if you like it commit. I have other ideas but will share them in Discord.

src/editor/editor_settings/editor_settings.go
```diff
@@ -91,6 +92,7 @@ func (s *Settings) setDefaults() {
 	s.RefreshRate = 60
 	s.CodeEditor = "code"
 	s.UIScrollSpeed = 20
+	s.ShowGrid = false
```
BrentFarris, line 95: I'll probably default this to true for the merge

BrentFarris (approved):
👍 Great, thank you! I think the `View` menu item will find more uses as we continue. I also think people would like to hide the grid if they want to do any demo screen recordings or anything like that.

src/editor/editor_settings/editor_settings.go
BrentFarris: Added in d094f774591622dd5068a1dc145fef2e3885336b

Colorfingers:
Super, glad it was useful... more contributions coming.  I don't know if you will like the floating panels but I prefer it over the static ones.  Two things I generally don't like about most 3D engines... pinned menu panels and you cannot just drop your model fully skinned on the stage.

Issue #721, opened 2026-04-28
Minimizing the window (Windows) leads to a fatal error


BrentFarris:

BrentFarris:
This issue has previously been resolved by @gitsomebit

PR #737, opened 2026-04-30
fix(window): fix crash when minimizing window

gitsomebit:
pr summary:
- fix crash when minimizing editor window
- add some error handling7logging
- add some comments

NOTE: the fix was basicly only that `device.destroyGlobalUniforms()` was called before checking `g.SwapChain.IsValid()`. I can remove the other changes, but I think they would be useful for future development


BrentFarris (approved):
Thank you for addressing this!

PR-ref: #721

BrentFarris:
👍 Thank you for looking into this. I couldn't reproduce the crash on my main local machine, I haven't tried my laptop yet though. Others have reported it, so this is very helpful.

PR #738, opened 2026-04-30
feat(file-drop): allow dropping folders in window (windows)

gitsomebit:
pr summary:
- follow-up to #729
- allows folder-drop (windows)

NOTE: dropping folders with multiple images (exp 5) is very slow. 
TODO: need to look at import logic (also ui freezes during import)


BrentFarris (approved):
👍 Thanks!

PR #739, opened 2026-04-30
fix(import): speed up png import

gitsomebit:
pr summary:
- follow-up to #738
- speeds up import for png files, but total import time for many assets still slow

### DIFF
--- a/src/editor/project/project_database/content_database/content_database_texture.go
+++ b/src/editor/project/project_database/content_database/content_database_texture.go
@@ -71,9 +71,22 @@ func (Texture) ExtNames() []string { return []string{".png", ".jpg", ".jpeg", ".
 func (Texture) Import(src string, _ *project_file_system.FileSystem) (ProcessedImport, error) {
 	defer tracing.NewRegion("Texture.Import").End()
 	var decoder func(r io.Reader) (image.Image, error) = nil
+
 	switch strings.ToLower(filepath.Ext(src)) {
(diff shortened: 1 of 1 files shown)


gitsomebit:
NOTE: this pr introduces issues when the png payload has invalid data because the texture is not validated at runtime and can cause crashes!

@BrentFarris do we want to add runtime validation when creating images or close this pr?

BrentFarris:
@gitsomebit I will probably close this one. I just tested out an idea for threading the import and it is working pretty good. I'll like the commit once I put it in, to this thread

BrentFarris:
Go flexing it's concurrency benefits here:
https://github.com/KaijuEngine/kaiju/commit/9e62619972ad15688d985417358c937ade9261d9

https://github.com/user-attachments/assets/1548b47e-bb7e-4144-bcee-0dc9c79b1668

BrentFarris:
I like the idea of the PR skipping PNG import as it's the native format, but it does come with a potential error as you described. So, we may attempt to improve some other algorithms to speed up import speeds before hopping onto some short cuts.

Thank you for the PR! It shed light on the problem and was inspirational to getting this improved. There is still some more we can improve here, but the above commit should be a good start.

BrentFarris:
Here's a better before/after

https://github.com/user-attachments/assets/17b6672c-c131-4f48-8bde-db3b54901374

PR #740, opened 2026-04-30
css sizing

gitsomebit:
pr summary:
- implement css sizing

use this file for testing visually
[css_sizing.html](https://github.com/user-attachments/files/27261384/css_sizing.html)
still looks a bit off here and there, but I think that is because of other not yet implemented properties

### DIFF
--- a/src/engine/ui/markup/css/properties/css_aspect_ratio.go
+++ b/src/engine/ui/markup/css/properties/css_aspect_ratio.go
@@ -37,7 +37,7 @@
 package properties
 
 import (
-	"errors"
+	"fmt"
 
 	"kaijuengine.com/engine"
 	"kaijuengine.com/engine/ui"
@@ -46,7 +46,23 @@ import (
 )
 
 func (p AspectRatio) Process(panel *ui.Panel, elm *document.Element, values []rules.PropertyValue, host *engine.Host) error {
-	problems := []error{errors.New("AspectRatio not implemented")}
+	if len(values) == 0 {
+		return fmt.Errorf("AspectRatio requires at least 1 value")
+	}
+	valueStrs := make([]string, 0, len(values))
+	for i := range values {
+		valueStrs = append(valueStrs, values[i].Str)
+	}
+	ratio, ok := parseRatio(valueStrs)
+	if !ok {
+		setAspectRatio(panel, 0, false)
+		return nil
+	}
+	setAspectRatio(panel, ratio, true)
 
-	return problems[0]
+	layout := panel.Base().Layout()
+	width := applyWidthConstraints(panel, layout.PixelSize().Width())
+	height := applyHeightConstraints(panel, width/ratio)
+	layout.Scale(width, height)
+	return nil
 }
--- a/src/engine/ui/markup/css/properties/css_aspect_ratio.go
+++ b/src/engine/ui/markup/css/properties/css_aspect_ratio.go
@@ -55,10 +55,10 @@ func (p AspectRatio) Process(panel *ui.Panel, elm *document.Element, values []ru
 	}
 	ratio, ok := parseRatio(valueStrs)
 	if !ok {
-		setAspectRatio(panel, 0, false)
+		disableAspectRatio(panel)
 		return nil
 	}
-	setAspectRatio(panel, ratio, true)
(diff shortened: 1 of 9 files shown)


BrentFarris (approved):
Tested it out, it's great progress. I'm happy to see min/max constraints come in, they are one of the things I do tend to use a lot but haven't been able to get around to.

Thanks! 👍

BrentFarris:
Update on this PR:

I was just testing out some things and realized that my main stage was getting pretty slow. It turned out that I had accidentally used `min-height` instead of `height` [in stage_workspace.go.html](https://github.com/KaijuEngine/kaiju/commit/340540999b48cdbb0eaf046751f5ce0f1781ea6b). The interface got pretty laggy, it seems like that was enough to trigger long UI stabilization loops (breaks out after a [certain number of iterations](https://github.com/KaijuEngine/kaiju/blob/master/src/engine/ui/ui.go#L299-L300)). So, there seems to be a case where the UI tree doesn't become stable in some cases when using `min-height`.

This has happened historically from floating point jitter or from one CSS rule clobbering another and causing the UI to dirty itself repeatedly. Sometimes setting the `Sort() int` in `css_property_types.go` helps fix this by ensuring certain properties happen before/after others. I'm not sure if it could solve this issue, but I'll keep looking around, maybe tomorrow, to see what's going on here.

BrentFarris:
Update, this looks to be due to the panel `FittingContent` specifically `ContentFitHeight`. This height can be less than the specified `min-height`, making it drop to the smaller height, then getting `min-height`'ed back to the min height.

My initial thoughts is that we store the `minSize matrix.Vec2` and `maxSize matrix.Vec2` on the `panelData` instead of storing it in entity named data. We can then say that there is a constraint on the min sizes if the component in question (z/y) of `minSize` is not `0`, and likewise for the `maxSize`. This will reduce some of the code in `css_sizing_constraints.go` and allow a panel's fit content code to be aware of the values and to respect them.

[The panel code in question](https://github.com/KaijuEngine/kaiju/blob/master/src/engine/ui/panel.go#L512-L526)

BrentFarris:
It's late, and I need to head to bed, but this is the basic idea for resolving the issue; let me know if you think it's reasonable (or if I missed something, I'm a bit tired). If not, we can revert it and take another swing at it.

https://github.com/KaijuEngine/kaiju/commit/d3b2d86b235fc6059ea2482c4784ccea60b5fb9f

gitsomebit:
> I was just testing out some things and realized that my main stage was getting pretty slow

sorry did not catch that in my testing, but slow how exactly? how do you measure it so I know what you mean

> Sometimes setting the Sort() int in css_property_types.go helps fix this by ensuring certain properties happen before/after others.

sounds like a baind aid for a deeper bleeding.

I havent had that much experience with layouting implementation yet, but other approaches I could think of on the spot are using a round/ceil/floor or use int instead

In the meantime its probably best to revert the change

gitsomebit:
> I was just testing out some things and realized that my main stage was getting pretty slow

sorry I did not catch that during my testing. how did you measure it or what was the giveaway?

> https://github.com/KaijuEngine/kaiju/commit/d3b2d86b235fc6059ea2482c4784ccea60b5fb9f

drawback is that min-height/width of 0 is ignored and should be documented somwhere, but I think its a reasonable tradeoff considering there are other prefered ways to hide an element

BrentFarris:
> sorry I did not catch that during my testing. how did you measure it or what was the giveaway?

Basically my FPS went to something like 5 FPS due to the UI trying to stabilize but never being able to. The UI continues to try and stabilize and gives up after too many tries to not stall the game entirely.

> drawback is that min-height/width of 0 is ignored and should be documented somwhere, but I think its a reasonable tradeoff considering there are other prefered ways to hide an element

This is true, I like the 0 initialize being an invalid value for min/max, but it does prevent collapsing in that way. We could use an invalid number (negative) on initialization of the panel if we want to support max-* collapses.

gitsomebit:
I would be fine with negative values being disabled min/max values

do you work on this?

BrentFarris:
> I would be fine with negative values being disabled min/max values
> 
> do you work on this?

I can add this change/update

PR #744, opened 2026-05-01
fix(input): fix input crash and add input types

gitsomebit:
pr summary:
- fix crash when parsing unknown input 
- add more input types

html file that would cause crash before:
[input.html](https://github.com/user-attachments/files/27280561/input.html)


BrentFarris (approved):
Great, thank you! 👍

PR #745, opened 2026-05-01
feat(css-background): implement css background support basics

gitsomebit:
pr summary:
- support css background color
- support css background image url (not tested because idk how kaiju syntax differs)

tested on:
[demo-file.html](https://github.com/user-attachments/files/27283866/demo-file.html)

NOTE: aspekt ratio element seems to have an issue with background, but that is not related to this pr (will figure that out later)


BrentFarris (approved):
👍 Thanks, good start!

PR #746, opened 2026-05-01
improve some css code formating

gitsomebit:
pr summary:
- add some whitespace
- fix some error comments


BrentFarris (approved):
👍 I'm okay with this. Typically I don't add white space in my functions, but for CSS parsing stuff, it can make it more readable since they are quite linear and have various branching based on params.

Thanks!

PR #747, opened 2026-05-01
fix(css): simplify if branches with switches

gitsomebit:
pr summary:
- replace if branching logic with switch in border color files
- no logic change


BrentFarris (approved):
The cleanup is appreciated, thank  you!

PR #749, opened 2026-05-01
fix(css): simplify padding Preprocess()

gitsomebit:
pr summary:
- remove `removeRule` variable
- add comment for TODO because the error handling fails silently

### DIFF
--- a/src/engine/ui/markup/css/properties/css_padding.go
+++ b/src/engine/ui/markup/css/properties/css_padding.go
@@ -64,6 +64,7 @@ func paddingSizeFromString(elm *document.Element, str string, idx matrix.VectorC
 	default:
 		size = helpers.NumFromLength(str, window)
 	}
+
 	current[idx] = size
 	return current, nil
 }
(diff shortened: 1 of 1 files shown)


src/engine/ui/markup/css/properties/css_padding.go
```diff
@@ -80,23 +81,18 @@ func (Padding) Preprocess(values []rules.PropertyValue, rules []rules.Rule) ([]r
...
-			removeRule = true
 		case "padding-bottom":
 			values[2] = rules[i].Values[0]
-			removeRule = true
 		case "padding-left":
 			values[3] = rules[i].Values[0]
-			removeRule = true
-		}
-		if removeRule {
+		default:
```
BrentFarris, line 95: This inverts the prior logic. The remove rule was happening whenever any of the above cases happened, as the conditional check was `if removeRule {`.

So, it would have been equivalent to something like:
```go
switch rules[i].Property {
    case "padding-top", "padding-right", "padding-bottom", "padding-left":
        rules = slices.Delete(rules, i, i+1)
        i--
    default:
        // Do not remove the rule
}
```

However, this removes the rule inversely:
```go
switch rules[i].Property {
    case "padding-top", "padding-right", "padding-bottom", "padding-left":
        // Do not remove the rule
    default:
        rules = slices.Delete(rules, i, i+1)
        i--
}
```

BrentFarris (changes requested):
Thank you for the PR! Requesting changes to double check the intent

gitsomebit:
agree. will close this

PR #750, opened 2026-05-01
fix(css-width): fix width access [0]

gitsomebit:
pr summary:
- css-width code was accessing values[0] before length check


BrentFarris (approved):
👍 Thank you!

PR #752, opened 2026-05-01
fix(css): reduce nesting in css code

gitsomebit:
pr summary:
- reduce code nesting for some css code
- no logic changes


BrentFarris (approved):
👍 Thank you!

src/engine/ui/markup/css/properties/css_color.go
```diff
@@ -59,26 +59,29 @@ func setChildTextColor(elm *document.Element, color matrix.Color) {
...
+	}
+
+	if newHex, ok := helpers.ColorMap[hex]; ok {
+		hex = newHex
+	}
+
+	color, err := matrix.ColorFromHexString(hex)
+	if err == nil {
+		return err
+	}
```
BrentFarris, line 78: Should be `if err != nil {`, I'll fix this in post.

gitsomebit: good catch!

PR #754, opened 2026-05-02
fix(css-bg): fix background to work for label

gitsomebit:
pr summary:
- make bg work with label
- fixes crash issue for #753


BrentFarris (approved):
👍 Thank you!

PR #761, opened 2026-05-02
fix(context-menu): reposition

gitsomebit:
pr summary:
- fix bug: when doing right click in on content in content workspace whlie context menu is already open it would nor reposition

demo:

https://github.com/user-attachments/assets/7333ad50-b7ed-4f2c-8063-dccecef01b2a


BrentFarris (approved):
👍 Awesome, thank you!

PR #763, opened 2026-05-02
fix(input): fix numKey input

gitsomebit:
pr summary:
- fix numKeys for input in ui preview workspace


BrentFarris (approved):
👍 Works great, thank you! I actually had this on my list of things to do because it started annoying me a bit too much with the transform input; so, this is very exciting to see.

PR #770, opened 2026-05-03
feat(import): imported file extension

gitsomebit:
pr summary:
- imported files get their file extension to enable syntax hightlighting when opening with editor

NOTE: during testing I realized that changing GUID can crash/deadlock the editor, but that was already before this pr. will look into that

### DIFF
--- a/src/editor/project/project_database/content_database/content_database.go
+++ b/src/editor/project/project_database/content_database/content_database.go
@@ -115,7 +115,7 @@ func Import(path string, fs *project_file_system.FileSystem, cache *Cache, linke
 	dependencyMap := map[string][]ImportResult{}
 	for i := range proc.Variants {
 		res[i].Category = cat
-		res[i].generateUniqueFileId(fs)
+		res[i].generateUniqueFileId(fs, filepath.Ext(path))
 		if useLinkedId && linkedId == "" {
 			linkedId = res[i].Id
 		}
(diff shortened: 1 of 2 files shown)


gitsomebit:
thx for solvig merge conflict. still needs merge into master 👍

BrentFarris:
> thx for solvig merge conflict. still needs merge into master 👍

Sorry for not commenting on this yesterday. I'm still considering this one a bit for long-term usability. The omission of the extension was due to potentially changing the file type in the future opaquely. Say, instead of using PNG as the internal type, it uses some other format. This would require all of the references to be updated to match the new file extension. One alternative I considered for this case would be to allow referencing the content by id (ignoring the extension), that way the underlying type (and extension) can be swapped out in the future without needing to update all references.

I'll keep on thinking through this one, I know its an inconvenience to have to set the text editor to the file type ever time you open it. Perhaps a virtual file can be created to operate on that will sync with the concrete file. File systems are a bit of a pain, but still necessary to cooperate with.

gitsomebit:
I understand what you mean, but extensions are at the end of the day only type hints. If at some point it was decided that the engine chooses to switch internal type defaults it would break the existing files eighter way

BrentFarris:
Using the extensions as type hints for the OS to present previews is valuable. I think tonight I'll look into making it possible to continue to pull content using just the id without requiring the extension as well.

gitsomebit:
> I think tonight I'll look into making it possible to continue to pull content using just the id without requiring the extension as well.

not fully understand this part.
In my mind how I would handle changing internal type represenataion is make some "update logic" for old project versions using the old type

PR #772, opened 2026-05-03
feat(uuid): change uuid to v7 for content

gitsomebit:
pr summary:
- change uuid to uuidV7

why:
because content in in content workspace is sorted by its guid by default. because v7 has a time component, this change will group new imports by default and list them in historic order (unless guid is modified)


BrentFarris (approved):
👍 Sounds good to me, thanks!

PR #773, opened 2026-05-03
fix(css): fix min/max width/height

gitsomebit:
pr summary:
- fix disabeling max width/height
- change so 0 is disabled min width/height


BrentFarris (approved):
👍  Woops, thank you!

PR #777, opened 2026-05-06
Widows file dialogue

gitsomebit:
pr summary:
- rewrite native file dialogue (windows)
- mode for opening folders
- custom dialogue fields (dropdowns and checkboxes)
- option to restric navigation to a root folder and subfolders
- make filedialogue optional with build tag `filedialogue`
- add documentation in docs with code snippets

details:
- dialog requests execute on a worker thread
- completed dialog results are queued and processed on the window polling thread
- if `Root` is set, folder navigation is blocked outside that root and accepted selections are validated before results are returned
- start folder rule: `CurrentDirectory` is used only when it is inside `Root`; otherwise the dialog starts from `Root`

NOTE: I included some demo code in `editor_game_interface.go` so you can test the dialogs when running the editor. that code should be removed before merge

NOTE: add missing tag for filedrop in run file and docs


BrentFarris:
Thanks! I'm looking through this as I have time

BrentFarris:
I'll come back to this tonight, sorry for the delay

gitsomebit:
@BrentFarris no worries, we are all busy

did some thinking. maybe we should not add the advanced stuff (advanced options and limiting selection). It adds an unproportional amount of code for functionality that will rarely be used, if not never. But its your call to make

BrentFarris:
> @BrentFarris no worries, we are all busy
> 
> did some thinking. maybe we should not add the advanced stuff (advanced options and limiting selection). It adds an unproportional amount of code for functionality that will rarely be used, if not never. But its your call to make

I'd be in favor of reducing the complexity (not doing the advanced stuff). It can always be added if it's absolutely necessary. I'm always in favor of starting off simple and building out what is actually needed/required.

Are you interested in removing that stuff for this branch/PR, or is that something you'd prefer that I stripped out myself?

gitsomebit:
are you fimiliar enough with the code to strip it? if so, go for it 👍 
can also look over if you want. your call

BrentFarris:
> are you fimiliar enough with the code to strip it? if so, go for it 👍 can also look over if you want. your call

I'm sure it's not too hard to pick up on what's going on and strip it myself. Though it might be a bit longer for me to get around to it. I've been working off my bucket list recently hah

PR #778, opened 2026-05-08
fix(mat3): incorrect row, column and multiply assign

gitsomebit:
pr summary:
- fix mat3 row, column and multiply assign
- add test to ensure correctness


BrentFarris:
Woops! I didn't see that you made this PR earlier and just so happened to update the same code while I was creating tests for all of the other math files

gitsomebit:
seems to be fixed now. will close

PR #781, opened 2026-05-08
fix(err): fix flipBook[0] invalid access

gitsomebit:
pr summary:
- fix flipBook[0] fallback that is reached when len(flipBook) == 0


BrentFarris (approved):
👍 Great, thanks!

PR #786, opened 2026-05-09
fix(doc-links): use proper comment links suported by go

gitsomebit:
pr summary:
- use official go link syntax in comments
- only comments updated


BrentFarris (approved):
👍 Great, thanks!

PR #787, opened 2026-05-09
fix(icon): fix icon blackness

### LINKED ISSUE #766: app icon becomes black sometimes (windows)
sometimes when running editor in debug mode the icon seems not to be set correctly and turns black:

| taskbar | window icon |
|----------|----------|
| Image |  Image |

not yet sure what this causes as it is semi random and i dont have a good way to reproduce yet. might be a timing issue

Platform: Windows

gitsomebit:
pr summary:
- fix #766
- fix leaking old icons when changing icons

detail:
I think the black icon issue was related to `CreateCompatibleBitmap` sometimes producing black icons. I am not fully sure why that is, but I replaced it with `CreateDIBSection` and everything seems to be fine now.
while at it I added better error handling and resource cleanup.

### DIFF
--- a/src/platform/windowing/win32.c
+++ b/src/platform/windowing/win32.c
@@ -1189,34 +1189,80 @@ void window_set_file_drop_enabled(void* hwnd, bool enabled) {
 #endif
 
 void window_set_icon(void* hwnd, int width, int height, const uint8_t* pixelData) {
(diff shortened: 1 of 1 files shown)


BrentFarris (approved):
👍 Thanks!

🤔 Interesting, Windows has a lot of history and therefore a lot of ways to do the same thing with varied success.

PR #798, opened 2026-05-11
Add UI & console log to example game

jacobsorme:
Thoughts on adding a little more to the beginner example? Personally think some UI basics could be useful, as well as some logging. Cheers


BrentFarris (approved):
👍 Sure thing, thanks!

Issue #791, opened 2026-05-10
Game debug console auto scroll bottom?


jacobsorme:
Would it be possible to have the debug console auto scroll to bottom? I'm on Mac, doesn't seem to be the case there.

E.g. these prints
`console.For(g.host).Write(cam.Position().String())`

jacobsorme:
Could we CSS this? E.g this one https://stackoverflow.com/questions/18614301/keep-overflow-div-scrolled-to-bottom-unless-user-scrolls-up

BrentFarris:
We still have `flex` CSS work to do in the engine, so it can't be resolved immediately through CSS, however, it should be simple enough to update the console UI Go code to scroll to bottom automatically when a new message comes in (if you're already at the bottom).

jacobsorme:
Gotcha, will take a look! 🔧

jacobsorme:
Kinda related to #493

PR #809, opened 2026-05-15
#791 Fix console scroll stick

jacobsorme:
Fix scroll sticking to bottom on scroll to bottom. Non-stick on scroll up again, and on typing e.g. `help` which has a tall output, larger than hotnum 50. 

Is there a way to get the line height if Panel? See there is one for Label. Could make this a bit less hotnummed

### DIFF
--- a/src/engine/systems/console/console.go
+++ b/src/engine/systems/console/console.go
@@ -256,6 +256,12 @@ func (c *Console) update(deltaTime float64) {
 	host := c.host.Value()
 	debug.EnsureNotNil(host)
 	kb := &host.Window.Keyboard
+
+	lblParent, ok := c.doc.GetElementById("consoleContent")
+	if ok && matrix.Abs(matrix.Float(lblParent.UIPanel.ScrollY())-lblParent.UIPanel.MaxScroll()[1]) < 50 {
+		lblParent.UIPanel.SetScrollY(matrix.FloatMax)
+	}
+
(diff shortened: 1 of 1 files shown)


jacobsorme:
See now that we have the `c.outputLabel()` but it doesn't have a line height set, because the console output is one label with a single multi row string? If we really want the line height integrated here maybe we could convert the output to many labels?

BrentFarris:
> See now that we have the `c.outputLabel()` but it doesn't have a line height set, because the console output is one label with a single multi row string? If we really want the line height integrated here maybe we could convert the output to many labels?

I'm fine with converting the console to multiple labels, that might be a better approach. We could then re-use labels as they go off-screen so that we could have a (basically) "limitless" amount of entries. That would basically make it so that there are only N labels total (enough to fit within the container), then as one goes out it can be moved/recycled to the bottom for the next label scrolling into view.

jacobsorme:
Gotcha, sounds good. Would also enable us to add individual colors on the labels/rows which could be nice. Make that in another PR and proceed with this if you're OK with the hotfix here?

BrentFarris:
> Gotcha, sounds good. Would also enable us to add individual colors on the labels/rows which could be nice. Make that in another PR and proceed with this if you're OK with the hotfix here?

That works, thanks! I think being able to color the lines in the console will be nice so that we can identify warnings/errors quickly.

BrentFarris (approved):

PR #811, opened 2026-05-17
Add obj import + logging example to sample game

jacobsorme:
Thoughts on adding examples of mesh import and logging to the sample game?


BrentFarris (approved):
👍 I think having some more basics like this in the example is great, thank you!

PR #818, opened 2026-05-20
feat: add support for mono spaced font for code-editor plugin

djgilcrease:
* Adds JetBrainsMono for FontMono for use in the code-editor plugin I am working on
* Adds `OnFileSaved` event that the code-editor plugin will use to trigger formatting
* Fixes an issue with the font color ranges I ran into
* Fixed a bug loading plugins that have their own deps
* Fixed a bug when trying to disable plugins due to how the recompile code would cache the plugin list
* Ensure plugins are recompiled after editor upgrade/rebuild
   * Prompts user to recompile or ignore for this session and disables the plugin for this session if ignored
* Ensure editor plugins are recompiled after plugin code change
   * Prompts user to recompile/ignore if ignored will just run the old plugin code as it was
* Fix loading of project if `.DS_Store` some how ended up in `database/config` directory

UI for a newly built editor with plugins enabled but not compiled in currently
image

UI if your plugin source has changed since you last recompiled the editor
image


BrentFarris (approved):
👍 Awesome, thanks!

PR #819, opened 2026-05-21
Feature/css cursors

### LINKED ISSUE #180: Implement all of the CSS cursors (css_cursor.go)
This needs the cursors to be added across all the platforms

claudioluciano:
adds native and virtual cursor support with themes (:

native cursors uses the cursor provided by each platform, falling back to the default arrow.
virtual cursors hides the platform and render a UI cursor from a selected theme.
add the Kenney cursor pack (CC0)

auto mode uses native cursor when available and fall back to virtual cursor otherwise.

closes #180


BrentFarris:
Thank you! It might take me a bit to get through this as I have time after work hours, but I'll get to this asap.

PR #820, opened 2026-05-22
Begin console individual labels, colors & reuse

jacobsorme:
See #791 & #809 

Begin try making console rows/entries individual labels, to:
- Support individual entry colors, style
- Possibly improve console scroll-stick-to-bottom logic

Current limitations:
- ~~`lbl.SetColor(matrix.ColorAquamarine())` isn't showing, unsure why. I've tried other approaches than using `.Clone` on an element, like building manually, but gotten stuck in nil pointer issues~~ 

> Fixed in 90cc0598dc0795c2da8f8340355dcb99d80cf9b4. Had to use `EnforceFGColor`, otherwise the entries got their color from parent (`setChildTextColor(elm, color)` override)

- Hard coded `1000` limit on the entries, no history of past

Thoughts:
- I guess it's possible (as discussed in #809) to keep a history (`[]string`?), detect scroll up at the top and then shift currently showing entries downward. However, would this make the scroll look "discrete" and not "continuous" (shift/scroll _whole_ rows, not fractions)?
- Would the scroll bar at the side reflect the real number of entries if we cap visually? Maybe that's not an issue?

Feedback welcome! 

image

### DIFF
--- a/src/engine/systems/console/console.go
+++ b/src/engine/systems/console/console.go
@@ -164,7 +164,7 @@ func (c *Console) ExecCommand(key, arg string) (string, error) {
 
 func (c *Console) Write(message string) {
 	lbl := c.outputLabel()
-	lbl.SetText(lbl.Text() + "\n" + message)
+	lbl.SetText(message)
 }
 
 func (c *Console) help(*engine.Host, string) string {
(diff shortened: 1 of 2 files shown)


jacobsorme:
Close, stale

PR #824, opened 2026-05-23
Move kb check before active check + cursor offset console

jacobsorme:
Console hotfix / bugfix


BrentFarris (approved):
👍 Thanks!

For context, the early return was added due to `lblParent.UIPanel.SetScrollY` firing every frame, even when the console was not visible.

jacobsorme:
Oh damn, good catch! Sorry about that :pray:

PR #826, opened 2026-05-24
Add oklab/oklch color + tests

jacobsorme:
Initially thought of adding this for CSS reasons: many CSS component libraries use `oklch` and Kaiju having it would bring usage of such components closer. 

However, the CSS syntax of `oklch(59.69% 0.156 49.77 / .5)` (with `%` and `/`) felt tricky to support ATM. Saving that for the future. For now adding general Oklab + Oklch support here in the `matrix` package. 

Oklab sources: 
1. Oklch to Oklab https://bottosson.github.io/posts/oklab/#the-oklab-color-space
2. Oklab to linear sRGB https://bottosson.github.io/posts/oklab/#converting-from-linear-srgb-to-oklab
3. Linear sRGB to display RGB https://bottosson.github.io/posts/colorwrong/#what-can-we-do%3F

See also
- https://oklch.com/#0.7,0.103,29.05,50
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/oklch
- https://tailwindcss.com/docs/color (Tailwind usage of `oklch`)

Run the tests with `go test -run TestColorOklab kaijuengine.com/matrix -v` 
 
Thoughts on adding this?

### DIFF
--- a/src/matrix/color.go
+++ b/src/matrix/color.go
@@ -85,6 +85,7 @@ func OklabToColor(l, a, b, alpha Float) Color {
 		} else {
 			c[i] = 12.92 * c[i]
 		}
+		c[i] = Max(Min(c[i], 1), 0) // Naive clamping, could instead reduce chroma until no overflow
 	}
 	return c
 }
--- a/src/matrix/color.go
+++ b/src/matrix/color.go
@@ -9,6 +9,7 @@ package matrix
 import (
 	"encoding/binary"
 	"fmt"
+	"math"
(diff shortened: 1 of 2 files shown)


BrentFarris (approved):
I think this is a good idea. I've not seen this Oklab stuff, honestly. However, having the availability of the functions is interesting, they are free-floating, so it's harmless to have them in place here. If you'd like, you can create an issue around this for the CSS implementation of it.

PR #828, opened 2026-05-24
change value to pointer receiver

gitsomebit:
pr summary:
- change progressbar value to pointer receiver

NOTE: I think it was a typo. ProgressBar is a big struct where pointer makes more sence


BrentFarris (approved):
👍 Correct, thank you!

PR #836, opened 2026-05-27
fix: flex layout fixes when using complex column and row based flex

djgilcrease:
Was running into issues with column flex, trying to vertically center, even when set to flex-start, as well as general viewport sizing issues caused by a divide by 0 when it was checking world size

### DIFF
--- a/src/engine/ui/markup/css/properties/css_overflow_y.go
+++ b/src/engine/ui/markup/css/properties/css_overflow_y.go
@@ -41,7 +41,6 @@ func (p OverflowY) Process(panel *ui.Panel, elm *document.Element, values []rule
 		panel.SetOverflow(ui.OverflowScroll)
 		panel.Base().GenerateScissor()
 		panel.DontFitContentHeight()
-		panel.DontFitContentHeight()
 	case "inherit":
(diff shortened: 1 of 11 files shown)


src/engine/ui/markup/css/properties/css_overflow_y.go
```diff
@@ -40,6 +40,8 @@ func (p OverflowY) Process(panel *ui.Panel, elm *document.Element, values []rule
 		panel.SetScrollDirection(panel.ScrollDirection() | ui.PanelScrollDirectionVertical)
 		panel.SetOverflow(ui.OverflowScroll)
 		panel.Base().GenerateScissor()
+		panel.DontFitContentHeight()
+		panel.DontFitContentHeight()
```
BrentFarris: Duplicated line

BrentFarris (changes requested):
Minor change request

src/engine/ui/markup/css/properties/css_overflow_y.go
djgilcrease: fixed

BrentFarris (approved):
Thanks!

Issue #835, opened 2026-05-27
Graviton mesh read error?


jacobsorme:
Testing out Graviton (crazy stuff!) in a non editor project, getting some `gob` errors on the meshes:

```go
{"time":"2026-05-27T21:57:52.331114+02:00","level":"INFO","msg":"read","verts":896,"indexes":1512}
{"time":"2026-05-27T21:57:52.332749+02:00","level":"INFO","msg":"read","verts":960,"indexes":1440}
{"time":"2026-05-27T21:57:52.339244+02:00","level":"ERROR","msg":"failed to read graviton mesh physics shape","assetKey":"hub.obj","error":"gob: bad data: undefined type arrayType = struct { CommonType CommonType = struct { Name string; Id int; }; Elem int; Len int; }"}
{"time":"2026-05-27T21:57:52.340512+02:00","level":"ERROR","msg":"failed to read graviton mesh physics shape","assetKey":"lock.obj","error":"gob: bad data: undefined type arrayType = struct { CommonType CommonType = struct { Name string; Id int; }; Elem int; Len int; }"}
```

The meshes show up correctly in the game, looking sane by the looks of it. Could there be something wrong with meshes, or something in the import? Been trying to look around for that `arrayType = struct { CommonType ... `. Got this one at first, by messing up the `AssetKey`, error above showed up after fixing name:

```go
{"time":"2026-05-27T21:55:40.40547+02:00","level":"INFO","msg":"read","verts":960,"indexes":1440}
{"time":"2026-05-27T21:55:40.409729+02:00","level":"ERROR","msg":"failed to read the mesh","id":"hub","error":"openat hub: no such file or directory"}
```

Loading the meshes: 

```go
meshCache := host.MeshCache()
res, _ := loaders.OBJ("hub.obj", host.AssetDatabase())
meshCache.Mesh("hub", res.Meshes[0].Verts, res.Meshes[0].Indexes)
```

`kaiju_mesh.go` seems to be where it's triggered:
```go
// Deserialize will construct a [KaijuMesh] from the given array of bytes. This
// supports the current native mesh format and falls back to gob for legacy
// assets.
func Deserialize(data []byte) (KaijuMesh, error) {
	defer tracing.NewRegion("kaiju_mesh.Deserialize").End()
	if isNativeMesh(data) {
		return deserializeNative(data)
	}
	r := bytes.NewReader(data)
	dec := gob.NewDecoder(r)
	var km KaijuMesh
[...]

PR #837, opened 2026-05-29
Change graviton init for mesh

jacobsorme:
Spent some more time on #835, had some problems with `ShapeMesh` in Graviton:

- Made some changes in `func (r RigidBodyEntityData) gravitonRigidBody` for mesh specifically, looked like the `ShapeMesh` were unconditionally set to static? In-game I couldn't get them moving.  
- Changed `SetStaticMesh` to `SetShapeMesh` (same but without the `SetStatic`part), and in above let `ShapeMesh` run through the same flow as the rest (static/dynamic setup)
- Aligned the tests (`go test kaijuengine.com/engine/graviton/ -v` OK)

I can verify in game that the `ShapeMesh` now move around, and other types can collide with them. I've not been able to get two `ShapeMesh` to collide with each other though.

No hurry looking through this at all, just thought I could add an update.

### DIFF
--- a/src/engine/graviton/mesh_test.go
+++ b/src/engine/graviton/mesh_test.go
@@ -33,7 +33,8 @@ func TestStaticMeshBodyGeneratesBroadPhaseAABB(t *testing.T) {
 	system.Initialize()
 	body := system.NewBody()
 	body.Transform.SetPosition(matrix.Vec3{3, 0, 0})
-	body.SetStaticMesh(testTriangleMesh())
+	body.SetShapeMesh(testTriangleMesh())
+	body.SetStatic()
 	system.broadPhase.Rebuild(&system.bodies)
 	if len(system.broadPhase.proxies) != 1 {
 		t.Fatalf("expected 1 broad phase proxy, got %d", len(system.broadPhase.proxies))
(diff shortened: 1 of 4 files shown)


jacobsorme:
> I've not been able to get two ShapeMesh to collide with each other though.

Found this, will take a look to see if I can find something that's already implemented and use that

```go
func collideBodyMeshPair(a, b *RigidBody) (Contact, bool) {
	if a.Collision.Shape.Type == ShapeTypeMesh || b.Collision.Shape.Type == ShapeTypeMesh {
		if a.Collision.Shape.Type == ShapeTypeMesh && b.Collision.Shape.Type == ShapeTypeMesh {
			return Contact{}, false
		}
```

Will two nested `ForEachWorldTriangle` be feasible for mesh-mesh? Use `concurrent.Threads` in there, or reserved only for running top level separate collision work in parallel?

jacobsorme:
_(Edit: Just FYI, not included in this PR)_
Made a naive attempt at mesh-mesh (`m1` to `m2`) by using the existing sphere-mesh (`collideSphereMesh`), essentially creating a sphere for each triangle in `m1` (using `m1.Bounds.Center` and the triangle centroid to get center and radius) and running sphere-mesh for each towards `m2`. Doesn't seem to work and also slow. 

Found this one https://en.wikipedia.org/wiki/Möller–Trumbore_intersection_algorithm. Something to use in sphere-mesh, and maybe generalize for mesh-mesh? Cheers

BrentFarris:
Hello! Thanks for the effort and information on this, sorry for the delay. I'll review some of these details as soon as I have a moment.

BrentFarris (approved):
👍 Thanks!

PR #846, opened 2026-06-03
fix: address cpu hotpath and css rendering issue

djgilcrease:
* enhance font support for transparent label and panel
* ensure text gets centered in its own panel
* switch to `strconv.ParseFloat` instead of `fmt.Scanf("%f", ...)`
* Allow having switchable themes in css with vars, they were statically resolved at parse before
* cache dpm instead of calling out to c every frame


BrentFarris (approved):
👍 Great UI updates, thank you very much!

PR #849, opened 2026-06-05
feat: add a virtualized panel, text doc, and refactor textarea

djgilcrease:
* Added VirtualList a vertically-scrolling list that only materializes the rows currently in view
* Added ability to show/hide the scrollbar `SetScrollbarsVisible`
* Added a `documentModel` line-based text buffer that works with the VirtualList and TextArea to handle fast rendering of massive amounts of text data and allows an undo/redo buffer history as well.
* Refactored TextArea to use VirtualList and documentModel to speed up rendering of large text blocks that need lots of coloring


BrentFarris (approved):
👍 Awesome, thank you!

PR #850, opened 2026-06-05
Add asset database for embed directive files

jacobsorme:
Thoughts on adding a database for `//go:embed` `embed.FS`? General thoughts on `embed` for assets in Golang game dev? 

Plan I had was to enable us doing something like:
```go
//go:embed editor/editor_embedded_content/editor_content game_content
var gameContent embed.FS
```
over in `main.test.go` (or maybe with more granular paths)

Btw, thought of modifying existing `FileDatabase` to make it `fs.FS`-based instead, hence making it support both current `os.Root` (with minor wrapping to make compatible) and `embed.FS`. Not sure if that's desirable tho, `os.Root` n `embed.FS` being stricter and all

### DIFF
--- a/src/engine/assets/embed_database.go
+++ b/src/engine/assets/embed_database.go
@@ -1,0 +1,94 @@
+/******************************************************************************/
+/* file_database.go                                                           */
+/******************************************************************************/
+/* MIT License, Copyright (c) 2015-present Brent Farris, (John 4:13-14)       */
+/******************************************************************************/
+
+package assets
+
+import (
+	"embed"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"kaijuengine.com/platform/profiler/tracing"
+)
+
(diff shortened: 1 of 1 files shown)


BrentFarris:
> Thoughts on adding a database for //go:embed embed.FS? General thoughts on embed for assets in Golang game dev?

I think it's an interesting angle to explore, basically single executable games. I actually like embedding content, it simplifies a lot of stuff, though there'd need to be some special magic if people intend to compress their assets in any way (if they want to).

BrentFarris (approved):
👍 Thanks!

Let's put it in and let people have the option for embedded databases. There is no cost here. I like giving developers the option to develop games how they want to and not enforce any one method. This is partly why I left the `AssetDatabase` open ended as an interface. Developers could even create one to pull from SQLite or a remote server (not a recommendation) haha.

Issue #720, opened 2026-04-28
Implement screen count on Linux and MacOS


BrentFarris:
This is from the PR:  #719 

There is a [TODO for Mac](https://github.com/KaijuEngine/kaiju/pull/719/changes#diff-b0f4a5f2df2cb1123370a00c8e3ca790625c17b4189ffbcdf1980bbf0e92930eR481).

There is a [TODO for Linux](https://github.com/KaijuEngine/kaiju/pull/719/changes#diff-ad6773353fc3e3bed8e6d66d72d9fc2bca30d9aaaf7b9eaf34524df99dee1062R644).

Remove [this note](https://github.com/KaijuEngine/kaiju/pull/719/changes#diff-56e5922593e2b72274ca3b7b136262e712cfb3b28a6db0beb52a10e99226947cR264) when complete.

PR #852, opened 2026-06-10
feat: implemented screen count in mac

relaxgameing:
with reference to: #720 

this is the implementation of the screen count for mac 

```
Note: 
* the code is generate by ai 
* it has not been tested for more than 1 monitor 
```

since it looks theoretically correct , i created a pull request 😅


BrentFarris (approved):
👍 Thanks!

PR #853, opened 2026-06-15
Add primitive shapes to mesh selection popup #814

### LINKED ISSUE #814: Add primitive shapes to mesh selection popup
When you click on the mesh select button on an entity, it should present primitive shapes like sphere, cylinder, etc. A list of the primitive shapes that should be selectable are in the Create menu.

Image

Image

Image

ghost:
When clicking the mesh select button on an entity, the popup now lists
  all primitive shapes available in the Create menu: Cube, Sphere, Plane,
  Capsule, Cylinder, Cone, and Arrow. The primitive entries are appended to the content selector options before
  "None" when the selector is opened for the Mesh type. Selecting a
  primitive calls the existing change Mesh path which already supported
  built in mesh IDs via rendering. BuiltInMeshData. The details panel also
  now displays the readable name for example "Sphere" instead of the raw
  internal key when a primitive mesh is set on an entity.

  Fixes #814

### DIFF
--- a/src/editor/editor_overlay/content_selector/content_selector_overlay.go
+++ b/src/editor/editor_overlay/content_selector/content_selector_overlay.go
@@ -18,6 +18,7 @@ import (
 	"kaijuengine.com/engine/ui/markup/document"
 	"kaijuengine.com/platform/hid"
 	"kaijuengine.com/platform/profiler/tracing"
+	"kaijuengine.com/rendering"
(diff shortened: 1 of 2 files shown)


BrentFarris:
Thank you! I'll take a look at this soon

BrentFarris (approved):
👍 Awesome, thank you!

PR #855, opened 2026-06-20
feat: generalize the virtual list

djgilcrease:
This generalizes the virtual list to allow custom layout calculations, which are needed to support a custom RecyclerView (modeled after the Android implementation) in my widget library. I only added the vertical layout to the core, so it will continue to work for the text area. I will implement horizontal, grid, flex, and Mason layouts in my widget library.


BrentFarris (approved):
👍 Thank you!

Issue #858, opened 2026-06-21
F64 build


jacobsorme:
It's been discussed before (see e.g. #806) that the `float64` build (`go build -tags="debug,F64" -o ../bin/kaiju`) needs some revision to get new changes aligned with `matrix.Float` being a `float64`. 

Made a static check on the number of errors you currently get when adding the `F64` build tag: 

> 1127 Total 
337 kaijuengine.com/engine/ui
143 kaijuengine.com/rendering
106 kaijuengine.com/engine/ui/markup/css/properties
77 kaijuengine.com/engine/cameras
65 kaijuengine.com/engine/systems/tweening
61 kaijuengine.com/editor/editor_stage_manager/editor_stage_view/transform_tools
50 kaijuengine.com/engine/systems/visual2d/sprite
45 kaijuengine.com/editor/editor_controls
34 kaijuengine.com/engine/graviton
33 kaijuengine.com/integration_testing
24 kaijuengine.com/editor/editor_stage_manager/editor_stage_view
23 kaijuengine.com/platform/hid
21 kaijuengine.com/editor/editor_workspace/stage_workspace
18 kaijuengine.com/editor/editor_overlay/color_picker
10 kaijuengine.com/rendering/vfx
9 kaijuengine.com/editor/editor_workspace/render_graph_workspace
8 kaijuengine.com/framework
8 kaijuengine.com/editor/editor_workspace/ui_workspace
6 kaijuengine.com/platform/windowing
6 kaijuengine.com/editor/editor_stage_manager/editor_stage_view/select_tool
6 kaijuengine.com/editor/editor_stage_manager/data_binding_renderer
4 kaijuengine.com/editor/editor_workspace/content_workspace
4 kaijuengine.com/editor/editor_overlay/context_menu
3 kaijuengine.com/rendering/loaders
3 kaijuengine.com/engine_entity_data/engine_entity_data_tween_transform
3 kaijuengine.com/engine_entity_data/engine_entity_data_skin_animation
3 kaijuengine.com/editor/project/project_database/content_previews
3 kaijuengine.com/editor/codegen/reflect_helpers
2 kaijuengine.com/platform/steam
2 kaijuengine.com/engine/ui/markup/css/functions
2 kaijuengine.com/engine/systems/console
2 kaijuengine.com/editor/editor_workspace/terrain_workspace
1 kaijuengine.com/engine/lighting
1 kaijuengine.com/editor/global_interface/status_bar
[...]

PR #859, opened 2026-06-21
Add matrix package float generics

jacobsorme:
See #858. Been thinking of a way to solve many `F64` build errors with minimal code changes, i.e. keeping struct `float32` fields in the `F64` build. 

 Thoughts on going for generics in some parts of the `matrix` package? 

Makes it so that many call sites work out of the box for both `float32` and `float64`. The `1127 Total` `F64` build errors (see #858) are reduced to: 

> 777 Total 
251 kaijuengine.com/engine/ui
104 kaijuengine.com/engine/ui/markup/css/properties
66 kaijuengine.com/rendering
...

Key change is this:

```go
func Min[T1, T2 tNumber](a T1, b T2) Float { // float32 args in float64 build OK
	return Float(math.Min(float64(a), float64(b)))
}
...
```

And

```go
func NewVec3[T1, T2, T3 tNumber](x T1, y T2, z T3) Vec3 { // Arg type can vary. Return val always according to build 
	return Vec3{Float(x), Float(y), Float(z)}
}
...
```

The former operations require no changes out in the code base to make it work for `F64`.

 The latter require changes from `matrix.Vec3{...}` to `matrix.NewVec3(...)` (some included in this PR) to be `F64` compatible.

The errors left after these changes seem to be mostly simple arithmetic with `float32` and `float64` mixups. Cheers!


BrentFarris (approved):
Using generics is a great solution to this problem, to be honest. Thank you!

Also thank you for the additional cleanup you've done in this PR.

PR #860, opened 2026-06-21
fix: make sure the x and y scroll is set properly

djgilcrease:
The trackpad horizontal scroll gesture was not working for two reasons
1. The deltaX was set on Y
2. w.Mouse.SetScroll is absolute so setting the X & Y needs to be made in a single call


BrentFarris (approved):
👍 Thanks!

PR #861, opened 2026-06-24
fix: fonts are index by codepoint

djgilcrease:
I tried to add a new font with a large set of symbol code points and this was causing it to not render the symbols


BrentFarris (approved):
👍 Thanks!

PR #863, opened 2026-07-01
feat: add an Apply to tweening that samples this easing curve at point t

djgilcrease:
Found this useful for a custom tween flow


BrentFarris (approved):
👍 Sweet, thank you!

Issue #790, opened 2026-05-10
[EDITOR] Camera snaps when rotating in 3D


Blengen:
On any project, in the 3D editor view, when rotating the camera using right-click, the camera snaps in the direction of the cursor away from the center

**To Reproduce**
1. Open any project
2. Enter 3D View
3. Right click the main window **whilst currently moving the mouse**
- Note: This only occurs when actively moving the mouse. If you right-click, and **then** move your mouse, it functions as expected.

**Likely cause**
Since the snap is proportional to the cursor's distance and direction from the screen center, the issue is likely that the engine believes you moved the cursor that far in the first frame.

**A simple solution** (if this is the case), is to simply ignore the first frame of rotation, letting the cursor first snap to the center.

**Platform**
OS: Linux Mint
Kaiju Version: Nightly (10th May 2026) - Windows Version (ran through Wine)

https://streamable.com/jba0sb

BrentFarris:
Thanks for the report! I believe your suggestion is a great suggestion that has minimal code change to solve the problem. I'll add it to the queue to fix.

relaxgameing:
Hi @Blengen , the video link is not working properly so i was not able to see what you meant.

is this the behaviour you are talking about

https://github.com/user-attachments/assets/43d86c8a-3ae2-49f0-b1b4-628c0f3941b5

Blengen:
@relaxgameing Sorry for the late answer. I couldn't upload the video directly for some reason, so I used Streamable.

As for the bug: Yes, it's the same bug, same root issue, but different cause (if my theory is correct)
Essentially, it tracks movement per frame by checking how far away the cursor is from the center. For me, it takes 1 frame to lock the cursor to the center, so if I move the cursor as I'm clicking, it's gonna turn the camera really far (snap).

Based on your video, the cursor just never snaps to the center at all, causing the same bug, but for as long as you are holding RMB. Could be a display server issue, not sure.

relaxgameing:
> As for the bug: Yes, it's the same bug, same root issue, but different cause (if my theory is correct) Essentially, it tracks movement per frame by checking how far away the cursor is from the center. For me, it takes 1 frame to lock the cursor to the center, so if I move the cursor as I'm clicking, it's gonna turn the camera really far (snap).
>
> Based on your video, the cursor just never snaps to the center at all, causing the same bug, but for as long as you are holding RMB. Could be a display server issue, not sure.

Hi @Blengen, thank you for confirming the issue. 
[...]

PR #864, opened 2026-07-01
fix: fly camera rotation works as expected #790

relaxgameing:
with Reference to: #790 

Hi @BrentFarris, Fly camera rotation now works as expected. No longer snaps

https://github.com/user-attachments/assets/927f70b9-b031-46e2-929d-78fba4a59848

Need your advice for the part commented on the [update function](https://github.com/relaxgameing/kaiju/blob/e40b26117a7630d543d0e1a023875a4ade5a9014/src/editor/editor_controls/editor_camera.go#L271-L287)


BrentFarris (approved):
👍 Awesome, thank you!

PR #866, opened 2026-07-02
feat: allow the particle system to be emitted in a volumetric space not just a point

djgilcrease:
Was adding a rain system and ran into oddities with camera movement culling all particles and look/feel needing to be more than just a single point emitter.


BrentFarris (approved):
👍 This is something I had on my do later list, so thank you!

PR #867, opened 2026-07-02
fix: osx would crash when loading a large object set with dynamic shadow generation

djgilcrease:
Loading in a few thousand objects and trying to live render them with shadows was causing a segfault. This fixed it, it now properly falls back to a white box default for the texture for a frame or two while the real texture is being loaded.


BrentFarris (approved):
👍 Great, thank you!

PR #868, opened 2026-07-02
feat: allow full unloading of texture/mesh

djgilcrease:
Found that when textures/meshes are scene/zone specific, they stay in memory, and I wanted a way to clear the old ones to manage memory better.


BrentFarris (approved):
👍 Thank you! I'll merge this in for now.

I'll probably think a bit on this one a bit more to determine if we need a "remove from cache" function that is a soft removal (previous logic), so the texture is evicted once all references are lost, and a "hard removal" which this is more akin to.

_If the previous soft removal wasn't working once all of the references were dropped, I may also need to look into that._

I also may also update the `CreatePending` function name to be something like `ProcessPending` as it is handling destruction.

PR #869, opened 2026-07-06
More F64 / cross-build prep

jacobsorme:
#### Sorry for lots of changes in the same PR, I think we've landed in a good solution now though. 
---

----
Took a stab at getting a complete `float32` and `float64` toolbox for `F64` build. 

- Made `Mat3f` and `Mat4f` `float32` logic for the `F64` build (hardcoded `float32`s instead of build dependent `Float`). Files `mat3.float64.go` and `mat4.float64.go`. Should be all but identical to the `float32` build variants
- Refactored the float32 ASM funcs to be available both as `Mat4Multiply` for `F32` build and as `Mat4fMultiply` for `F64`

Both regular and `F64` build compiles and runs (the `F64` just showing a blank window, still)


jacobsorme:
#858

src/matrix/mat4.go
```diff
@@ -492,36 +514,36 @@ func (m *Mat4) Inverse() {
...
-	m01 := m[4]
-	m11 := m[5]
-	m21 := m[6]
-	m02 := m[8]
-	m12 := m[9]
-	m22 := m[10]
+func (m Mat4T[T]) ToQuaternion() Quaternion {
+	m00 := Float(m[0])
+	m10 := Float(m[1])
+	m20 := Float(m[2])
```
jacobsorme: Dodged making `Quaternion` generic at this stage, convert to `Float` instead

src/matrix/float.go
```diff
@@ -86,3 +86,79 @@ func Round[T tNumber](x T) T {
...
+
+func Vec4MultiplyMat4T[T tFloatingPoint](v Vec4T[T], m Mat4T[T]) Vec4T[T] {
+	if _, ok := any(*new(T)).(float32); ok {
+		return any(Vec4fMultiplyMat4f(any(v).(Vec4f), any(m).(Mat4f))).(Vec4T[T])
+	}
+	return vec4MultiplyMat4Fallback(v, m)
+}
+
+func mat4MultiplyFallback[T tFloatingPoint](a, b Mat4T[T]) Mat4T[T] {
+	var result Mat4T[T]
```
jacobsorme: generic `mat4MultiplyFallback` used here as stand-in for `float64`, and if non arm or amd also as a stand-in for the otherwise ASM:ed `Vec4fMultiplyMat4f`

src/matrix/float.go
```diff
@@ -86,3 +86,79 @@ func Round[T tNumber](x T) T {
...
+
+type Vec4f = Vec4
+type Mat4f = Mat4
+
+func Mat4Multiply(a, b Mat4) Mat4 {
+	return Mat4MultiplyT(a, b)
+}
+
+func Mat4MultiplyT[T tFloatingPoint](a, b Mat4T[T]) Mat4T[T] {
+	if _, ok := any(*new(T)).(float32); ok {
```
jacobsorme: Realise we could prob check type on `a`, instead of directly on `T`. Will take another look later today hopefully

src/matrix/matrix.simd.go
```diff
@@ -9,10 +9,10 @@
 package matrix
 
 //go:noescape
-func Mat4Multiply(a, b Mat4) Mat4
+func Mat4fMultiply(a, b Mat4) Mat4
```
jacobsorme: Should prob change type to `Mat4f` here instead? Thought I did

BrentFarris:
Doing a bit of OT at work this week, I'll review this asap, thanks!

jacobsorme:
No worries, good luck! Just thought of reflect for new types, like you fixed after an earlier PR, maybe we’ll need a fix here too?

src/matrix/mat3.float64.go
```diff
@@ -0,0 +1,266 @@
...
+func Mat3fFromSlice(a []float32) Mat3f {
+	return Mat3f{
+		a[0], a[1], a[2],
+		a[3], a[4], a[5],
+		a[6], a[7], a[8],
+	}
+}
+
+func Mat3fFromVec3f(v Vec3f) Mat3f {
+	return Mat3f{
```
jacobsorme, line 34: Funcs like these probably don't need to be in a `F64` specific file.

Instead, have it shared across both builds and in `F32` build just add aliases `var Mat3FromSlice = Mat3fFromSlice`.

jacobsorme: Actually I think we can keep them as is. We need the `Float` funcs that'll work for both `F32` and `F64`. And then we need `float32` specific funcs for the `float64` build. 

We could opt for shuffling some code around but I don't think it would save us any LOCs.

BrentFarris: Yeah, the fact that we have to keep the `float32` types for working with Vulkan complicate this a bit, we can make the floating functions generic, but the type members throw things off if/when they are used together.

jacobsorme: Yea true. A key throw-off here seems to be the fact that we can’t both:

1) Have an instantiated type inherit all the generic ones defined on the generic type
2) And have custom methods defined later on the instantiated type itself.

We’ll get that error `cant define methods on instantiated type`.

This makes the generics a bit all or nothing, which isnt ideal for our special float handling.

However - instead of assigning a type by instantiating a generic one we can wrap it in a struct! That would give that new struct access to generic methods via the member, and we can define new methods on it. Wrapping in a struct feels a bit convoluted imo tho.

jacobsorme:
Close, stale

PR #870, opened 2026-07-09
Fix colors to float32

jacobsorme:
Thoughts on fixing `matrix.Color` to `float32`? Thinking it could save us from handling another color type, `Colorf` (`float32`fixed). Tried making colors generic, but proved a bit tricky with the special handling of `Color8` (255)

### DIFF
--- a/src/matrix/color.go
+++ b/src/matrix/color.go
@@ -13,22 +13,22 @@ import (
 	"strings"
 )
 
-type Color Vec4
+type Color Vec4T[float32]
 type Color8 [4]uint8
 
-func (c Color) R() Float                           { return c[R] }
-func (c Color) G() Float                           { return c[G] }
-func (c Color) B() Float                           { return c[B] }
-func (c Color) A() Float                           { return c[A] }
-func (c *Color) PR() *Float                        { return &c[R] }
-func (c *Color) PG() *Float                        { return &c[G] }
-func (c *Color) PB() *Float                        { return &c[B] }
-func (c *Color) PA() *Float                        { return &c[A] }
-func (c Color) RGBA() (Float, Float, Float, Float) { return c[R], c[G], c[B], c[A] }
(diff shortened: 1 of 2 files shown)


jacobsorme:
Or is replacing Float with float32 a slippery slope lol?

BrentFarris:
My overtime has relented, so I can finally catch up a bit here.

I actually was thinking the same thing, however, I am conflicted, because I like `Color` and `Vec4` effectively being the same underlying type. So I didn't pull the trigger on `float32`. It doesn't make a whole lot of sense to be a `float64`, as that precision doesn't make a whole lot of sense with gaming display infrastructure and would lean more into scientific realms more than game realms.

This would simplify a few points of the graphics code, but maybe complicate Vec4<->Color crossover. Perhaps I just need to figure out the surface area of Vec4<->Color, there likely isn't really much.

jacobsorme:
No worries, welcome back. This is `Vec4` ATM

```go
type Vec4 = Vec4T[Float]
``` 
Changing `Color` from `Vec4` to `Vec4T[float32]` shouldn't be make a difference in the `F32` build right? Maybe we can add `ToColor` method to `Vec4` and `ColorToVec4T` func to `color` package – is that the surface area? We don't seem to have any conversion helpers between the two currently. 

Btw initially I thought we had `type Color = Vec4` but we actually have `type Color Vec4`, meaning `Color` don't inherit any methods of `Vec4` ATM.

BrentFarris:
> Changing `Color` from `Vec4` to `Vec4T[float32]` shouldn't be make a difference in the `F32` build right? Maybe we can add `ToColor` method to `Vec4` and `ColorToVec4T` func to `color` package – is that the surface area? We don't seem to have any conversion helpers between the two currently.

Let's give this a shot, it's not exactly hard to pivot later if we find some fatal flaw in implementation.

BrentFarris (approved):

PR #871, opened 2026-07-15
fix: do not panic if buffer size is 0

djgilcrease:


BrentFarris (approved):
👍 Thanks! I do intend pretty much all of the calls to `panic` to get replaced at some point, so this is great.

PR #872, opened 2026-07-19
`float32`->`matrix.Float` replace

jacobsorme:
Made some curated replacements of `float32`->`matrix.Float`. If it's too much changes to merge for a PR I can send you the bash scripts and the list of packages. All builds below OK. 

```
go build -tags="debug,editor,F64" -o ../bin/kaiju
go build -tags="debug,editor" -o ../bin/kaiju
go build -tags="debug,F64" -o ../bin/kaiju
go build -tags="debug" -o ../bin/kaiju
```

One thing, do we have to align some reflection code with the type changes here? Saw the changes you made to `func CanonicalTypeName(typeName string) string {` to handle the generic types


BrentFarris:
Hmm, yeah, I looked through a bit of this and the main concern I have with the blanket update is when it comes to serialized fields. The entity data is serialized (reflected) to a binary serialization, so that one is higher priority to have as a fixed-size data type. The ones that serialize to JSON (like editor settings) is less concerning, but maybe would be nice to be fixed-size for future portability.

jacobsorme:
Hmm, entity data is that editor only? 

Thinking if we can just skip float32->matrix.Float replace for all editor things

BrentFarris:
> Hmm, entity data is that editor only?
> 
> Thinking if we can just skip float32->matrix.Float replace for all editor things

Unfortunately, entity data is serialized by the editor and used in the running game instance. It is what allows you to setup data in your game, expose it to the editor UI for modification, then have that data serialized (with the stage) for the game to load at runtime.

jacobsorme:
Close, stale

PR #882, opened 2026-08-31
Prevent editor from freezing during project load and added loading overlay

thvgger:
• Asynchronous Loading: Moved ed.project.Open(), ed.project.Initialize(), and ed.project.TryUpgrade() inside editor_project_setup.go into background Goroutines.
  • Main Thread Synchronization: Once the heavy file operations and parsing are complete, execution is handed back to the UI thread via ed.host. RunOnMainThread() to safely finalize the engine state and UI setup.
  • Loading UI Indicator: Added a clean, non-intrusive loading modal (.loadingModal) to the new_project_overlay.go.html template. It safely blocks interactions (like duplicate clicks on "Create Project") by using an invisible .loadingOverlay shield that covers the panel until the background Go routines invoke the close() callback.

### DIFF
--- a/src/editor/editor.go
+++ b/src/editor/editor.go
@@ -228,58 +228,65 @@ func (ed *Editor) effectiveRefreshRate(status platformPower.Status) int32 {
 	return klib.Clamp(refreshRate, 0, 320)
 }
 
-func (ed *Editor) postProjectLoad() {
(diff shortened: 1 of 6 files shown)


BrentFarris:
👍 Awesome, thanks!

Issue #885, opened 2026-09-05
OBJ loader causes panic on empty line


loissascha:
**Describe the bug**
I noticed that trying to load a .obj file that I have exported from Trenchbroom causes a panic because the exported file contains empty lines. 

**To Reproduce**
1. Create a new map in Trenchbroom
2. Export it as .obj
3. Try to import it in your game using loaders.OBJ

**Expected behavior**
Empty lines should be ignored

**Additional context**
Seems like in obj_loader.go it tries to access the 0 index of the line it's trying to parse, but if the line is empty it causes a index out of range error.

PR #886, opened 2026-09-05
Prevent panic from empty line in .obj file

loissascha:
Hello. This is my first contribution and I hope I'm following all the right rules. 

I've reported the issue #885 which I fixed in this pull request. I've also added a test that covers the reported issue with empty lines.

When a .obj file has an empty line in it it runs into an index out of range error and panics because it tries to access runes[0] in the obj_loader


BrentFarris (approved):
Awesome, this is great, thank you!

